PR #8, opened 2022-10-06
add missing fragments

cong-or:
_explorer_update_proposal_certificate_test()_  covers _**update_proposal**_ and _**update_vote**_


saibatizoku (approved):
Looks good to me. Only thing that I don't entirely get is why the test should panic, perhaps a comment in the test would be good for future dev.

kukkok3 (approved):
LGTM

PR #11, opened 2022-10-10
Add Cargo manifest for monorepo

saibatizoku:
* Creates top-level Cargo.toml that declares a workspace
* Moves workspace members into top-level
  * Most of the profile settings for legacy workspaces are put into the individual crates in their manifest, and at the moment are being ignored by Cargo.
  * Removes legacy workspaces by removing the Cargo manifest that defines them.

- [X] Still needs some cleanup of inner Cargo.lock files


cameron1024 (approved):
Looks good to me, but probably want to wait for other team members to have a look before merging

cameron1024:
I guess we also want to think about how this impacts the nix build. IIRC some of the crates have their own workspace-aware nix build scripts, and we probably don't want to break those

saibatizoku:
> I guess we also want to think about how this impacts the nix build. IIRC some of the crates have their own workspace-aware nix build scripts, and we probably don't want to break those

Yeah, I'm doing it because the nix flakes were having problems, mainly (e.g. flake file needs to be on same directory as Cargo.lock). I guess that the next step is to refactor nix for the monorepo.

cong-or:
LGTM

garbas:
> > I guess we also want to think about how this impacts the nix build. IIRC some of the crates have their own workspace-aware nix build scripts, and we probably don't want to break those
> 
> Yeah, I'm doing it because the nix flakes were having problems, mainly (e.g. flake file needs to be on same directory as Cargo.lock). I guess that the next step is to refactor nix for the monorepo.

Correct, this will make nix packaging much easier.

Mr-Leshiy (approved):

PR #12, opened 2022-10-10
Expose protocol update voting in REST Api

dkijania:
Exposed updates/active endpoint for taking out currently active update proposals. Extended one of automation tests which should cover basic functionality

### DIFF
--- a/src/jormungandr/jormungandr-lib/src/interfaces/updates.rs
+++ b/src/jormungandr/jormungandr-lib/src/interfaces/updates.rs
@@ -1,0 +1,48 @@
+use crate::interfaces::{
+    config_params::FromConfigParamError, BlockDate, ConfigParams, ConsensusLeaderId,
+};
+use chain_impl_mockchain::{certificate::UpdateProposal, update::UpdateProposalState};
+use serde::{Deserialize, Serialize};
+
+#[derive(Deserialize, Serialize, Debug)]
+pub struct UpdateProposalStateDef {
+    pub proposal: UpdateProposalDef,
+    pub proposal_date: BlockDate,
+    pub votes: Vec<ConsensusLeaderId>,
+}
+
+impl TryFrom<UpdateProposalState> for UpdateProposalStateDef {
+    type Error = FromConfigParamError;
+    fn try_from(state: UpdateProposalState) -> Result<Self, FromConfigParamError> {
+        Ok(UpdateProposalStateDef {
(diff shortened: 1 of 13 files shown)


src/jormungandr/jormungandr-lib/src/interfaces/updates.rs
```diff
@@ -0,0 +1,48 @@
...
+impl TryFrom<UpdateProposalState> for UpdateProposalStateDef {
+    type Error = FromConfigParamError;
+    fn try_from(state: UpdateProposalState) -> Result<Self, FromConfigParamError> {
+        Ok(UpdateProposalStateDef {
+            proposal: state.proposal.try_into()?,
+            proposal_date: state.proposal_date.into(),
+            votes: state
+                .votes
+                .into_iter()
+                .map(|(x, _)| ConsensusLeaderId::from(x.clone()))
```
Mr-Leshiy, line 23: Are you sure that it is needed `x.clone()` here, seems it should work just with `x`.

src/jormungandr/jormungandr-lib/src/interfaces/updates.rs
```diff
@@ -0,0 +1,48 @@
...
+#[derive(Deserialize, Serialize, Debug)]
+pub struct UpdateProposalDef {
+    pub config_params: ConfigParams,
+    pub proposer_id: ConsensusLeaderId,
+}
+
+impl TryFrom<UpdateProposal> for UpdateProposalDef {
+    type Error = FromConfigParamError;
+    fn try_from(update_proposal: UpdateProposal) -> Result<Self, FromConfigParamError> {
+        let mut config_params = Vec::new();
```
Mr-Leshiy, line 38: I am not sure that it will work.
Seems that it would be better to get `update_proposal.proposer_id()` first
```
let proposer_id = ConsensusLeaderId::from(update_proposal.proposer_id().clone());
```
So then config_params could be created in the following way to remove additional clonning

```
for config_param in update_proposal.changes().into_inter() {
             config_params.push(config_param.try_into()?);
         }
```

src/jormungandr/jormungandr-lib/src/interfaces/updates.rs
dkijania: unfortunately into_iter impl for Hamt return iter for inner collection

Mr-Leshiy (approved):

PR #13, opened 2022-10-10
remove unsafe send impls

cameron1024:
Removes a few unsafe `Send` impls in chain-libs

A quick look doesn't reveal any unsoundness, since there were other `Send` bounds in places where they should be, but it's a much more fragile solution than relying on auto traits to get the implementation

Also fixes a few clippy errors


minikin (approved):
LGTM

>but it's a much more fragile solution than relying on auto traits to get the implementation

@cameron1024 Should we create a task for refactoring?

cameron1024:
We don't need to add this manually, the auto trait stuff happens automatically, we just used to have an unsafe impl to get around a slightly confusing error message. But since it's not strictly needed, we shouldn't use it

PR #19, opened 2022-10-11
Attempt to fix vitup mock memory leaks problem

dkijania:
properly reset context in rest api. Fix dockers to use main version of catalyst-core


minikin:
@kukkok3 @nicopado Can you review this PR, please?

kukkok3 (approved):

PR #23, opened 2022-10-11
Npg 3334 public tally fix

cong-or:


src/jormungandr/testing/jormungandr-automation/src/jormungandr/explorer/verifiers/vote_plan_verifier.rs
```diff
@@ -72,6 +72,9 @@ impl ExplorerVerifier {
             );
             match &vote_proposal_status.tally {
                 Tally::Public { result } => {
+                    if explorer_proposal.tally.is_none() {
+                        continue;
+                    }
```
kukkok3, line 77: PublicTally cannot be None. Before the beginning of the tally time it will be all zeros `results: [0, 0, 0]` You can check the response of the voteplanstatus api So I would change this line of code with `assert!(explorer_proposal.tally.is_some())`

src/jormungandr/testing/jormungandr-integration-tests/src/jormungandr/explorer/vote_plan.rs
```diff
@@ -753,12 +749,10 @@ pub fn explorer_all_vote_plans_public_flow_test() {
...
-        .with_funds(
-            voters
-                .iter()
-                .map(|x| x.to_initial_fund(INITIAL_TREASURY))
-                .collect(),
-        )
+        .with_funds(vec![
+            voters[0].to_initial_fund(INITIAL_FUND_PER_WALLET_1),
+            voters[1].to_initial_fund(INITIAL_FUND_PER_WALLET_2),
+        ])
```
kukkok3, line 755: same as above

src/jormungandr/testing/jormungandr-automation/src/jormungandr/explorer/verifiers/all_vote_plans_verifier.rs
```diff
@@ -78,6 +78,9 @@ impl ExplorerVerifier {
                         );
                         match &vote_proposal_status.tally {
                             Tally::Public { result } => {
+                                if explorer_proposal.tally.is_none() {
+                                    continue;
+                                }
```
kukkok3, line 83: Same as below

src/jormungandr/testing/jormungandr-automation/src/jormungandr/explorer/verifiers/vote_plan_verifier.rs
cong-or: the unwrap fails [here](https://github.com/input-output-hk/catalyst-core/blob/npg-3334-public-tally-fix/src/jormungandr/testing/jormungandr-automation/src/jormungandr/explorer/verifiers/vote_plan_verifier.rs#L79) without the continue. It seems the [explorer_vote_plan ](https://github.com/input-output-hk/catalyst-core/blob/npg-3334-public-tally-fix/src/jormungandr/testing/jormungandr-automation/src/jormungandr/explorer/verifiers/vote_plan_verifier.rs#L12) stores the public tally as None at origin. Is this incorrect behaviour?

src/jormungandr/explorer/src/db/indexing.rs
```diff
@@ -479,10 +479,15 @@ impl ExplorerTransaction {
...
-    pub fn to_single_account(&self) -> Option<Identifier> {
+    pub fn to_identifier(&self) -> Option<Identifier> {
         match self {
             ExplorerAddress::New(address) => match address.kind() {
+                // Single address : A simple spending key. This doesn't have any stake in the system
                 chain_addr::Kind::Single(key) => Some(key.clone().into()),
+                // Account address : An account key. The account is its own stake
+                chain_addr::Kind::Account(key) => Some(key.clone().into()),
+                // Group address : an ed25519 spending public key followed by a group public key used for staking
+                chain_addr::Kind::Group(_spend, group) => Some(group.clone().into()),
```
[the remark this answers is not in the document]
cong-or: I was going by **[this](https://github.com/input-output-hk/catalyst-core/blob/main/src/chain-libs/chain-impl-mockchain/src/stake/distribution.rs#L204)** where the account key is equal to the identifier, I have named it group key here.

src/jormungandr/testing/jormungandr-integration-tests/src/jormungandr/explorer/vote_plan.rs
```diff
@@ -148,12 +147,10 @@ pub fn explorer_vote_plan_public_flow_test() {
...
```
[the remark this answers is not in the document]
cong-or: just to clarify, at the moment; the tokens equal funds in the voting centre? And tokens provide voting power. So, the initial funds here represent tokens and voting power.

I added _INITIAL_TOKEN_PER_WALLET_1_ for clarity - not sure if this is what you meant.

src/jormungandr/testing/jormungandr-automation/src/jormungandr/explorer/verifiers/vote_plan_verifier.rs
kukkok3: In my view yes it is incorrect. If you check the rest apis, in case the public tally is not started yet we return an array of 0 and the explorer should do the same to keep all our responses coherent with each other

src/jormungandr/testing/jormungandr-integration-tests/src/jormungandr/explorer/vote_plan.rs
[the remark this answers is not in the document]
kukkok3: Yes, we give tokens based on how many funds someone has. But this is an external rule, not enforced in the node. Here in principle we can assign as much tokens as we want, so I would keep two different values for the two different things to be sure we dont use one value for the other (has it was happening at one point)

What I meant is that having 
`to_initial_token(INITIAL_TOKENS_PER_WALLET_1)`
is more clear than
` to_initial_token(INITIAL_FUNDS_PER_WALLET_1),`

kukkok3 (approved):

PR #27, opened 2022-10-13
Fix explorer process incomplete logs in integration tests

FelipeRosa:
[ExplorerProcess](https://github.com/input-output-hk/catalyst-core/blob/426dc2fd7dce89d0943a4433805c0ebf74ed28f1/src/jormungandr/testing/jormungandr-automation/src/jormungandr/explorer/mod.rs#L54) logs were being only partially captured by the stdout pipe and written to `/tmp/jormungandr_*/explorer.log`. I'm assuming this is a buffering problem. This PR changes the Explorer's child process spawning code to capture its output to a file which seems to fix that problem.

I also did some refactoring.

### DIFF
--- a/src/jormungandr/testing/jormungandr-automation/src/jormungandr/explorer/mod.rs
+++ b/src/jormungandr/testing/jormungandr-automation/src/jormungandr/explorer/mod.rs
@@ -13,6 +13,7 @@ use crate::testing::configuration::get_explorer_app;
 use graphql_client::{GraphQLQuery, *};
 use jormungandr_lib::{crypto::hash::Hash, interfaces::BlockDate};
 use std::{
+    fs::File,
     process::{Command, Stdio},
     str::FromStr,
     time::Duration,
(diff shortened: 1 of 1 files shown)


src/jormungandr/testing/jormungandr-automation/src/jormungandr/explorer/mod.rs
```diff
@@ -176,24 +180,8 @@ impl ExplorerProcess {
...
-        } else {
-            return;
-        };
-
-        if std::thread::panicking() {
-            if let Some(logs_dir) = &self.configuration.logs_dir {
-                println!(
-                    "persisting explorer logs after panic: {}",
-                    logs_dir.display()
-                );
```
FelipeRosa, line 191: This was not displaying the path the logs would end up at. I believe the temporary file is moved to `/tmp/jormungandr_*/explorer.log` after it's written so it seemed ok for this to be removed (got confused by this log some times).

cong-or (approved):
LGTM. Works for me locally.

src/jormungandr/testing/jormungandr-automation/src/jormungandr/explorer/mod.rs
kukkok3: What do you think about catching the stderr as well like we do here https://github.com/input-output-hk/catalyst-core/blob/cf46276b599c26594c6c5f563f5d02fb0d00c69c/src/jormungandr/testing/jormungandr-integration-tests/src/jormungandr/explorer/explorer_sanity.rs#L203 ?

kukkok3 (approved):
LGTM

nicopado (approved):

src/jormungandr/testing/jormungandr-automation/src/jormungandr/explorer/mod.rs
FelipeRosa: Sure. Nice catch

PR #28, opened 2022-10-14
NPG-3845 Refactor logging mod and add OpenTelemetry layer to Explorer

FelipeRosa:
I made these changes while diving deeper at the explorer's code. The idea is to have a nice way of seeing tracing information to make debugging easier.

At the moment I added the OTLP exporter which can be used to export trace data to the [OTLP colletor](https://opentelemetry.io/docs/collector/) (or any other compatible backend). I used [this example](https://github.com/grafana/tempo/tree/main/example/docker-compose/otel-collector) to spin up a collector that uses [Grafana Tempo](https://grafana.com/docs/tempo/latest/) to store and query trace data through Grafana's UI.

image


saibatizoku:
This looks awesome!

cong-or:
Looks great!

cong-or (approved):

minikin:
@FelipeRosa Can we merge this one?

saibatizoku (approved):

src/jormungandr/explorer/src/logging.rs
```diff
@@ -129,66 +129,83 @@ impl LogSettings {
...
-        Ok((guards, self.msgs))
+        Ok(LogGuard {
+            _nonblocking_worker_guard: nonblocking_worker_guard,
+        })
     }
 
-    fn init_subscriber(&self, non_blocking: tracing_appender::non_blocking::NonBlocking) {
-        match self.config.format {
-            LogFormat::Default | LogFormat::Plain => {
+    fn init_subscriber(
```
cong-or, line 173: does this subscriber have global scope? just wondering how we can log everything in the explorer?

src/jormungandr/explorer/src/logging.rs
```diff
@@ -1,122 +1,122 @@
...
-        // Worker guards that need to be held on to.
-        let mut guards = Vec::new();
+impl Default for LogOutput {
+    fn default() -> Self {
+        Self::Stderr
+    }
+}
 
+pub struct LogGuard {
+    _nonblocking_worker_guard: tracing_appender::non_blocking::WorkerGuard,
```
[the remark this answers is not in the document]
FelipeRosa: From what I could understand, the `tracing_appender` that is being using to write spans has this `non_blocking` writer that spawns a separate thread for buffering and executing writes. When this guard is dropped, buffered spans are flushed and the thread is terminated. So we need to hold it for as long as we are emitting spans.

src/jormungandr/explorer/src/logging.rs
FelipeRosa: Yes. the `.init()` calls on the tracing subscriber `Registry` tries to set it as the global subscriber (https://docs.rs/tracing-subscriber/latest/tracing_subscriber/registry/struct.Registry.html#method.init).

FelipeRosa:
> recent changes look good +1
> 
> Only comment in relation to logging/telemetry out of the box.
> 
> do we have to add spans etc or is it good to go as is?

@saibatizoku this PR only sets up the functionally for exporting tracing events to an OpenTelemetry Collector. We still need to work on refactoring/improving the span structure and coverage.

PS: Exporting is disabled if no Collector endpoint is set while starting the explorer so things should work as before in that way.

PR #32, opened 2022-10-18
Implement new /api/v0/challenges/{id}/{voter_group_id} endpoint

Mr-Leshiy:
Adding new `/api/v0/challenges/{id}/{voter_group_id}` endpoint which acts mostly the same as `/api/v0/challenges/{id}` but also filtering proposals by `voter_group_id`

### DIFF
--- a/src/vit-servicing-station/doc/api/v0.yaml
+++ b/src/vit-servicing-station/doc/api/v0.yaml
@@ -205,6 +205,36 @@ paths:
         "404":
           description: The requested challenge was not found
 
+  /api/v0/challenges/{id}/{voter_group_id}:
+    get:
+      operationId: getChallenge
+      summary: Get challenge by id
+      tags: [challenge]
+      description: |
+        Retrieves information on the identified challenge,
+        including the proposals submitted for it filtered by the provided voter group id.
+      parameters:
+        - in: path
+          name: id
+          schema:
+            type: integer
+          required: true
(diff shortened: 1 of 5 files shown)


src/vit-servicing-station/doc/api/v0.yaml
```diff
@@ -205,6 +205,36 @@ paths:
         "404":
           description: The requested challenge was not found
 
+  /api/v0/challenges/{id}/{voter_group_id}:
```
cong-or, line 208: do we usually add [tests ](https://github.com/input-output-hk/catalyst-core/blob/main/src/vit-servicing-station/vit-servicing-station-tests/src/tests/rest/challenges.rs)for new endpoints? or is it redundant?

cong-or (approved):
LGTM. Just one comment about tests.

src/vit-servicing-station/doc/api/v0.yaml
dkijania: +1 @cong-or we should definitely satisfy decent automation coverage and since this is new endpoint we should add at least happy path for ti

dkijania (changes requested):
Please add at least on functional test or plan it in other PR

Mr-Leshiy:
@dkijania agreed, will do it in the separate PR. Will create a ticket for testing all endpoints from the "challenges" group

PR #34, opened 2022-10-19
[Refactor] Remove duplication in testing services - introduce scheduler lib

dkijania:
Added new project scheduler-service-lib which contains a toolkit to create simple scheduler service. Currently we have 3 implementations of scheduler service: 
- registration service (for working over voting-tools and invoke registartion transaction)
- snapshot service (like below but for fire up snapshot)
- registartion verify service (for validating registration transaction and qr code)

Thanks to that project we can remove duplication across mentioned projects


minikin:
@kukkok3 @nicopado Can you review it, please?

src/vit-testing/registration-service/src/client/args.rs
```diff
@@ -72,29 +67,13 @@ impl JobCommand {
                 Ok(())
             }
             Self::Status(status_command) => {
-                println!("{:?}", status_command.exec(rest)?);
+                println!("{:?}", status_command.exec::<State>(rest.into())?);
```
nicopado, line 70: debugging println?

nicopado (approved):

src/vit-testing/registration-service/src/client/args.rs
dkijania: actually not. this is what we print on to user since it's cli

PR #41, opened 2022-10-25
Add simple soak test for mock and protect its execution by soak feature

dkijania:


minikin:
@cameron1024 @Mr-Leshiy  review, please.

src/vit-testing/vitup/tests/mock/mod.rs
```diff
@@ -1 +1,13 @@
+use std::io::Write;
+use std::path::Path;
+use vitup::mode::mock::Configuration;
+
+#[cfg(feature = "soak")]
+mod soak;
 mod startup;
+
+pub fn write_config<P: AsRef<Path>>(config: &Configuration, output: P) {
```
Mr-Leshiy, line 9: Maybe return an errors here instead of unwrapping.

dkijania: it's only for component test so i would say that it's better than it's unwrapping because on  any issues test will panic sooner that later

Mr-Leshiy (approved):

PR #45, opened 2022-10-26
Npg 3808 tally per vote cast

cong-or:
- Update the public tally every time a vote is cast. Rather than waiting for the voting to end before being able to check tally results; mainly for _transparency_. 

- The rest of the PR addresses private tally return values in the None case _(3369)._


cong-or:
as discussed with Steven; an array of 0s is sufficient to represent the encrypted data until the voting and tally is complete.

kukkok3 (approved):

PR #47, opened 2022-10-31
Voting tools functional test

dkijania:
Added new functional test for voting tools. This change requires some small refacting of logic.rs module and introduction of DbProvider trait which provided abstraction over how data is accessed. Also i noticed potential for small refactoring in name convention for tests (reference and db-sync IMHO where wrongly placed names). 

After those changes i noticed that a big portion of mainnet_tools project can be removed or simplify if we expose votingtools test api and perform snapshot in memory thanks to previous changes


src/vit-testing/mainnet-lib/src/wallet/registration.rs
```diff
@@ -0,0 +1,190 @@
...
+        metadata.insert(
+            &BigNum::from(61285u32),
+            &TransactionMetadatum::new_map(&meta_sign_map),
+        );
+
+        metadata
+    }
+}
+
+pub trait GeneralTransactionMetadataInfo {
```
cameron1024, line 90: I wonder about having a trait vs just a struct here. What's the advantage of this approach vs having a function `get_metadata_info(&self) -> GeneralTransactionMetadataInfo` on types that we need it for?

src/vit-testing/mainnet-tools/src/bin/fake_snapshot_tool.rs
```diff
@@ -0,0 +1,47 @@
+use clap::Parser;
```
cameron1024, line 1: I'm not sure we want a separate binary for the tool with a mock data source. We don't duplicate too much, since most of the logic is in the voting tools `lib.rs`, but IMO it's cleaner and "closer to production" if we just have a CLI arg like `--mock-json-path` or something that is mutually exclusive with a set of real DB credentials

So I'm thinking a rough structure like this:
```rust
trait Db {
  // queries
}

struct RealDb { ... }
impl Db for RealDb { ... }

struct MockDb { ... }  // currently called `DbSyncInstance` IIRC
impl Db for MockDb {}

enum DbArgs {
  Real {
    host: DbHost,
    // ...
  },
  Mock {
    json_path: PathBuf,
  }
}

fn get_db(args: &DbArgs) -> Result<Box<dyn Db>> {
  // build the correct database here
}
```
I haven't finished looking at everything but this seems like a slightly cleaner approach. But there might be a reason this doesn't work, not sure :shrug:

src/voting-tools-rs/src/testing/test_api/fake.rs
```diff
@@ -0,0 +1,154 @@
...
+
+impl DataProvider for MockDbProvider {
+    fn vote_registrations(
+        &self,
+        lower: Option<SlotNo>,
+        upper: Option<SlotNo>,
+    ) -> color_eyre::Result<Vec<Reg>> {
+        let lower = if let Some(lower) = lower {
+            #[allow(clippy::cast_sign_loss)]
+            Some(lower.into_i64()? as u64)
```
cameron1024, line 30: We also probably want to use `try_into` here as well

src/voting-tools-rs/src/testing/test_api/fake.rs
```diff
@@ -0,0 +1,154 @@
...
+        } else {
+            None
+        };
+
+        let upper = if let Some(upper) = upper {
+            #[allow(clippy::cast_sign_loss)]
+            Some(upper.into_i64()? as u64)
```
cameron1024, line 37: `try_into` here too

src/voting-tools-rs/src/testing/test_api/fake.rs
```diff
@@ -0,0 +1,154 @@
...
+
+        let mut tx_id = 0;
+
+        Ok(filtered_regs
+            .iter()
+            .map(|(_block0, registrations)| {
+                registrations
+                    .iter()
+                    .map(|r| {
+                        let metadata = r.get(&TransactionMetadatumLabel::from(61284u32)).unwrap();
```
cameron1024, line 54: Maybe we could pull the `TransactionMetadataumLabel`s into consts (or statics if it's not possible to make them const)?

src/voting-tools-rs/src/testing/test_api/output_assertions.rs
```diff
@@ -0,0 +1,70 @@
...
+impl VerifiableSnapshotOutput for Output {
+    fn assert(&self) -> SnapshotOutputAssert {
+        SnapshotOutputAssert::new(self)
+    }
+}
+
+/// Fluent api for [Output] assertions
+pub struct SnapshotOutputAssert<'a> {
+    output: &'a Output,
+}
```
cameron1024, line 23: This seems like quite a lot of infrastructure for asserting stuff. Does this do something that a simple `assert_eq!()` in the test code wouldn't do?

cameron1024 (commented):
Looks good, nice to have an abstraction over mainnet data. 

I guess my general feedback is that we probably don't need so many traits for things. Abstracting over the database backend sounds like a good place, but I think a lot of other places would be made simpler by just using a single struct

src/vit-testing/mainnet-lib/src/db_sync/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+            tx_metadata: HashMap::new(),
+            stakes: HashMap::new(),
+            settings: Settings::default(),
+            db: temp_dir.child("db_sync.db").path().to_path_buf(),
+        }
+    }
+
+    /// Path to fake db = json file
+    #[must_use]
+    pub fn db_path(&self) -> &PathBuf {
```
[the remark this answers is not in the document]
dkijania: done

src/vit-testing/mainnet-lib/src/db_sync/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+use std::io::Write;
+use std::path::{Path, PathBuf};
+
+const CARDANO_MAINNET_SLOTS_PER_EPOCH: u64 = 43200;
+
+/// Mock of real cardano db sync. At this moment we only stores transactions metadata
+/// as the only purpose of existance for this struct is to provide catalyst voting registrations
+/// Struct can be persisted and restored from json file using `serde_json`.
+#[derive(Debug, Serialize, Deserialize)]
+pub struct DbSyncInstance {
```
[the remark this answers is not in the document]
dkijania: good idea. Renaming it to InMemoryDbSync since it's something bigger than mock

src/vit-testing/mainnet-lib/src/network/mod.rs
```diff
@@ -0,0 +1,194 @@
+use crate::db_sync::DbSyncInstance;
+use crate::wallet::MainnetWallet;
+use assert_fs::TempDir;
+use cardano_serialization_lib::address::Address;
+use cardano_serialization_lib::metadata::GeneralTransactionMetadata;
+use jormungandr_lib::crypto::account::Identifier;
+use jormungandr_lib::interfaces::BlockDate;
+use std::collections::HashSet;
+
+/// Mainnet newtor mock. It holds current block date as indicator on which epoch and slot mainnet
```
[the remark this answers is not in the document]
dkijania: done

src/vit-testing/mainnet-lib/src/wallet/mod.rs
```diff
@@ -0,0 +1,118 @@
...
+}
+
+impl MainnetWallet {
+    /// Creates new wallet with given ada. Currently wallet is purely used for testing purposes,
+    /// therefore we treat stake as arbitrary number not connected to any blockchain state.
+    /// # Panics
+    ///
+    /// Panics on key generation error
+    #[must_use]
+    pub fn new(stake: u64) -> Self {
```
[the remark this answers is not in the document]
dkijania: done

src/vit-testing/mainnet-lib/src/wallet/registration.rs
dkijania: I must confess that i introduced a trait for uncertain future because i feel like in other places in code (for example when testing registration transaction correctness etc.) would be handy. Currently yes, it looks like over engineering (however i think it encapsulated GeneralTrancationMetadata and registration builder doesn't know too much about it)

src/vit-testing/mainnet-tools/src/bin/fake_snapshot_tool.rs
dkijania: Happy that you proposed this, as i was uncertain if adding this approach won't be too close to production and can be misused, but i agree with additional parameter with meaningful name should solve the problem

src/voting-tools-rs/src/testing/test_api/fake.rs
dkijania: done

src/voting-tools-rs/src/testing/test_api/fake.rs
dkijania: done

src/voting-tools-rs/src/testing/test_api/output_assertions.rs
dkijania: Yes, again this is investment for the future as even now i have plan to add more test cases so this will avoid duplicating those assertions in other part of the repo

src/vit-testing/mainnet-lib/src/wallet/registration.rs
cameron1024: My usual strategy would be to introduce the trait in the same PR that we introduce the second implementation (in case we never get round to it, or decide against it, etc).

But it's not a huge deal, happy to go with your judgement

src/voting-tools-rs/src/testing/test_api/output_assertions.rs
cameron1024: Similar to above, I usually introduce traits/abstractions when they're needed by the code, but happy to go with this

dkijania: ok, thanks!

kukkok3 (approved):

src/voting-tools-rs/src/testing/test_api/output_assertions.rs
dkijania: Hm, i think that reviewer should resolve each conversation so that he accept my explanation or fix. But yes i can do it too ;)

PR #48, opened 2022-11-01
Javascript tests for wallet-js wasm bindings

Mr-Leshiy:
Implement a Javascript tests for the wallet-js wasm bindings.
Mainly this tests duplicates the same one which we have in rust, but in the Javascript and Nodejs.


minikin (approved):
LGTM

PR #52, opened 2022-11-04
Npg 2869 tx by addr query fix

cong-or:
fix transactionsByAddress query glitch which hangs when querying for inputs, outputs or certificate


kukkok3 (approved):
LGTM

PR #53, opened 2022-11-04
NPG-3383 Add Postgres support to vit-ss

FelipeRosa:
### Separate migration directories
I have separated the migration files for Postgres and SQLite because of difference in SQL types they have available.

So, for example, initializing a SQLite db can be done by: 
*(I'll add these examples to the docs later)*
```sh
diesel migration run --database-url /tmp/a.db --migration-dir vit-servicing-station-lib/migrations/sqlite
```

And for Postgres:
```
diesel migration run --database-url postgres://postgres:123456@localhost --migration-dir vit-servicing-station-lib/migrations/postgres
```

Both of them should map corretly to Diesel SQL types representations, e.g. Postgres `BYTEA` and SQLite `BLOB` map to diesel `Binary` SQL type.

### Db connection type is now an enum
Query support differs in both Postgres and SQLite. For example, Postgres doesn't have  `REPLACE INTO` queries so a `DELETE` followed by `INSERT INTO` should be used instead.

To be able to distinguish between `diesel::pg::PgConnection` and `diesel::sqlite::SqliteConnection` the db connection type is now a enum. Also because of type stuff.

### New macros
Sometimes we need to run the exact same query independent of the connection type being used (`PgConnection` or `SqliteConnection`) and Diesel will make them have different types. Diesel query traits bind the connection type to a database backend.

To avoid having to keep doing
```rust
match conn {
    DbConnection::Sqlite(inner) => query,
    DbConnection::Postgres(inner) => query,
}
```

There are a couple of new macros for that.

### Selecting the test database
Tests can be executed with a env var `TEST_DB` to select which database will be used. e.g.:

```
TEST_DB=postgres cargo test -p vit-servicing-station-lib
```

For now, Postgres URL is fixed.


src/vit-servicing-station/vit-servicing-station-lib/src/db/queries/challenges.rs
```diff
@@ -58,10 +68,13 @@ pub async fn query_challenge_proposals_by_id(
 ) -> Result<Vec<Proposal>, HandleError> {
     let db_conn = pool.get().map_err(HandleError::DatabaseError)?;
     tokio::task::spawn_blocking(move || {
-        proposals_dsl::full_proposals_info
-            .filter(proposals_dsl::challenge_id.eq(id))
-            .load::<Proposal>(&db_conn)
-            .map_err(|_e| HandleError::NotFound("Error loading challenge".to_string()))
+        q!(
```
cong-or, line 71: where is this q defined?

src/vit-servicing-station/vit-servicing-station-lib/src/db/models/challenges.rs
```diff
@@ -132,13 +140,13 @@ pub mod test {
...
-    pub fn populate_db_with_challenge_conn(
-        challenge: &Challenge,
-        connection: &PooledConnection<ConnectionManager<DbConnection>>,
-    ) {
-        diesel::insert_into(challenges::table)
-            .values(challenge.clone().values())
-            .execute(connection)
-            .unwrap();
+    pub fn populate_db_with_challenge_conn(challenge: &Challenge, connection: &DbConnection) {
+        q!(
```
cong-or, line 144: is it possible to test the q macro or is it infallible?

src/vit-servicing-station/vit-servicing-station-lib/src/db/queries/funds.rs
```diff
@@ -84,15 +99,20 @@ pub struct FundNextInfo {
...
-            .order(fund_dsl::id)
-            .limit(2)
-            .load(&db_conn)
-            .map_err(|_e| HandleError::NotFound("fund".to_string()))?;
+        let db_conn = &db_conn;
+
+        let funds: Vec<Fund> = q!(
+            db_conn,
+            fund_dsl::funds
+                // TODO: Not sure if sorting by the PK is actually necessary
```
cong-or, line 107: off topic; it would be nice to start removing these TODOs if they are no longer relevant.

FelipeRosa: I agree. Didn't touch them because I'm still not that familiar with the code and don't know whether they can be removed.

cong-or (approved):

src/vit-servicing-station/vit-servicing-station-lib/src/db/models/challenges.rs
FelipeRosa: I don't know if I got what you mean but It shouldn't fail at generating the code. Errors like that would be caught at compile time. What can be done, I believe, is test if it generates the expected code.

src/vit-servicing-station/vit-servicing-station-lib/src/db/mod.rs
```diff
@@ -28,12 +68,18 @@ WHERE
...
+    } else {
+        let manager = ConnectionManager::<SqliteConnection>::new(db_url);
+        let pool = Pool::builder().build(manager)?;
 
-    // test db connection or bubble up error
-    let conn = pool.get()?;
-    conn.execute(TEST_CONN_QUERY)?;
+        // test db connection or bubble up error
+        let conn = pool.get()?;
+        conn.execute(SQLITE_TEST_CONN_QUERY)?;
```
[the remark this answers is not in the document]
FelipeRosa: That code was already there. I didn't check if this is necessary. For Postgres an error is thrown if establishing the connection fails. I'll check this. Maybe it can be removed.

I tried to change the SQLite part of the code as little as possible to get vit-ss working with Postgres that's why there are things like this.

src/vit-servicing-station/vit-servicing-station-lib/src/db/models/challenges.rs
cong-or: ok cool, sorry yes it should be caught at compile time. no worries.

PR #55, opened 2022-11-08
ci from scratch

cameron1024:


cameron1024:
Caching is working, unfortunatey github actions is slow, takes 10m to run clippy even with a cache :(

flake.nix
```diff
@@ -278,35 +218,27 @@
...
+            openssl
+            protobuf
+            uniffi-bindgen
+            postgresql
+            diesel-cli
+            cargo-insta # snapshot testing lib
+          ];
           shellHook = ''
             echo "=== Catalyst Core development shell ==="
+            echo "Info: Git hooks can be installed using \`pre-commit install\`"
```
saibatizoku, line 237: This is no longer used.

flake.nix
```diff
@@ -278,35 +218,27 @@
...
           '';
-          # TODO: is this needed for vit-testing development
-          # export PATH="${jormungandr}/bin:$PATH"
-          # export PATH="${vit-servicing-station-server}/bin:$PATH"
         };
 
+        defaultPackage = warnToUpdateNix packages.default;
```
saibatizoku, line 241: Monorepo has no default package defined. This no longer applies.

saibatizoku (approved):
Looks good, thanks for the CI!

.github/workflows/ci_tests.yml
```diff
@@ -0,0 +1,71 @@
...
+
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+      - uses: Swatinem/rust-cache@v2
+      - name: Install deps
+        run:
+          sudo apt install -y protobuf-compiler libssl-dev libpq-dev libsqlite3-dev pkg-config
+      - name: Clippy and fmt
```
cong-or, line 66: should we add _-- -D warnings_ for clippy somewhere?

cameron1024: I believe that's the default behaviour when running `cargo clippy` but I'll double check when I'm at my computer

FelipeRosa (approved):

.github/workflows/ci_tests.yml
cameron1024: Yep, `cargo clippy` gives a non-zero exit code if any issues are found

PR #57, opened 2022-11-08
Archiving vit-ss database data into csv files

Mr-Leshiy:


src/catalyst-toolbox/catalyst-toolbox/src/archive/vit_ss.rs
```diff
@@ -0,0 +1,133 @@
...
+    #[serde(alias = "resultsUrl")]
+    pub results_url: String,
+    #[serde(alias = "surveyUrl")]
+    pub survey_url: String,
+}
+
+fn csv_writer(output_dir: &Path, name: String) -> Result<Writer<File>, Error> {
+    let mut path = output_dir.to_path_buf();
+    path.push(name);
+    let file = std::fs::File::create(path).unwrap();
```
FelipeRosa, line 75: This can return `Error` instead:

```suggestion
    let file = std::fs::File::create(path).map_err(Error::Io)?;
```

src/vit-servicing-station/vit-servicing-station-lib/src/db/queries/snapshot.rs
```diff
@@ -44,6 +44,21 @@ pub fn put_snapshot(snapshot: Snapshot, pool: &DbConnectionPool) -> Result<(), H
...
+pub async fn query_voters_by_snapshot_tag(
+    tag: String,
+    pool: &DbConnectionPool,
+) -> Result<Vec<Voter>, HandleError> {
+    let db_conn = pool.get().map_err(HandleError::DatabaseError)?;
+    tokio::task::spawn_blocking(move || {
+        voters::dsl::voters
+            .filter(voters::dsl::snapshot_tag.eq(tag))
+            .load(&db_conn)
+            .map_err(|e| HandleError::NotFound(format!("Error loading voters: {}", e)))
```
FelipeRosa, line 56: This will have to be changed to support both SQLite and Postgres backends, like:
```suggestion
        crate::q!(db_conn, voters::dsl::voters
            .filter(voters::dsl::snapshot_tag.eq(tag))
            .load(&db_conn))
            .map_err(|e| HandleError::NotFound(format!("Error loading voters: {}", e)))
```

src/vit-servicing-station/vit-servicing-station-lib/src/db/queries/snapshot.rs
```diff
@@ -106,6 +121,21 @@ pub async fn query_contributions_by_voting_key_and_voter_group_and_snapshot_tag(
...
+pub async fn query_contributions_by_snapshot_tag(
+    tag: String,
+    pool: &DbConnectionPool,
+) -> Result<Vec<Contribution>, HandleError> {
+    let db_conn = pool.get().map_err(HandleError::DatabaseError)?;
+    tokio::task::spawn_blocking(move || {
+        contributions::dsl::contributions
+            .filter(contributions::dsl::snapshot_tag.eq(tag))
+            .load(&db_conn)
+            .map_err(|e| HandleError::NotFound(format!("Error loading contributions: {}", e)))
```
FelipeRosa, line 133: Same as above

src/vit-servicing-station/vit-servicing-station-lib/src/db/queries/funds.rs
```diff
@@ -28,23 +28,34 @@ fn join_fund(
     fund.chain_vote_plans = voteplans_dsl::voteplans
         .filter(voteplans_dsl::fund_id.eq(id))
         .load::<Voteplan>(db_conn)
-        .map_err(|_e| HandleError::NotFound("Error loading voteplans".to_string()))?;
+        .map_err(|e| {
+            HandleError::NotFound(format!("Error loading voteplans, error: {}", e.to_string()))
+        })?;
```
FelipeRosa, line 33: This file will conflict with some changes I made to support both Postgres and SQLite databases. I'm happy to help resolve it if you like :)

FelipeRosa (changes requested):
A few comments

src/catalyst-toolbox/catalyst-toolbox/src/archive/vit_ss.rs
```diff
@@ -0,0 +1,133 @@
...
+
+    #[error(transparent)]
+    Io(#[from] std::io::Error),
+
+    #[error(transparent)]
+    Csv(#[from] csv::Error),
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq)]
+pub struct Fund {
```
dkijania, line 35: Vit servicing station project has it's own subproject which is dealing with loading csv into database:

https://github.com/input-output-hk/catalyst-core/blob/main/src/vit-servicing-station/vit-servicing-station-cli/src/csv/loaders.rs#L155

I wonder if this code could be a part of it. We can add another command for vit-servicing-station-cli command line tool for exporting data as csv. Keeping code here will somehow break SPOR rule IMHO and it's not logical (e.g. why i should use vit-servicing-station-cli for importing and catalyst-toolbox for exporting).

However this will introduce another problem that for user experience it would be better to use single cli for exporting everything rather than two (catalyst-toolbox for node, vit-ss-cli for vit-ss). Therefore my proposal is to  push code to vit-ss-cli project, expose it as lib and then reuse logic here. Thanks to that we wont duplicate responsibilities and the code

src/catalyst-toolbox/catalyst-toolbox/src/archive/vit_ss.rs
Mr-Leshiy: fixed

FelipeRosa (approved):
LGTM

src/catalyst-toolbox/catalyst-toolbox/src/archive/vit_ss.rs
Mr-Leshiy: I think personally all this cli tools should be a part of the catalyst-toolbox and takes in one place, especially as we have moved to the monorepo. So better I would move vat-servicing-cli as a part of the catalyst-toolbox, so as a result collect all our cli tooling in one place

dkijania: My opinion is not very strong but i would like to share different perspective.

I think we should go more into single responsibility for cli and then join some of the functionalities in catalyst-toolbox as i perceive it as big back of "don't know where to put it? throw it to catalyst-toolbox".  For example we have also `jcli` which is a node cli. Imagine that i just want to send a transaction to the node, why i need to reach catalyst-toolbox if i want to do so, i need to know that there is some global uber tool that can do it.  

The same situation is with vit-ss-cli, if i want to generate token for vit-ss which is the very narrow operation. I need to know that this functionality is in catalyst-toolbox which doesn't seems natural. Natural would be to perform it from vit-ss-cli. 

Catalyst-toolbox should be for operations that are on the upper level. Like i want to archive entire production env (nodes + vit-ss) then i can do it via single commands (jcli, vit-ss-cli ... by the way i think that archive node operation should be in jcli :) for the same reason as above) if I know what i'm doing. But if i want to do global archiving then catalyst toolbox command can be build on top of jcli and vit-ss-cli command, to satisfy this requirement Just like linux commands, we don`t have one uber cli tool which can do everything but smaller ones which can be piped etc. Here entire code is in one technology, so we can do this `piping` on api level.

Looking forwards to good discussion :)

Mr-Leshiy: Hmm, I see what do you mean. And I agree with you.
I am also don't have a strong position here :) 
For me personally I am logically separate all our cli's in the following way:
- cli which is needed to interact with jormungandr or provide something to jormungandr, it should be in `jcli`
- cli which facilitates something **for** vit-ss or interacting with vit-ss, it should be inside vit-ss-cli
- some general stuff should be inside catalyst-toolbox
So I am defining this feature as a _general stuff_, that is why I have put it into the catalyst-toolbox

minikin: @Mr-Leshiy @dkijania What's next here? 🤷

Mr-Leshiy: I would leave as it is, and merge it

minikin: Agree.

PR #59, opened 2022-11-08
Update nix flake

saibatizoku:
Sets Rust stable v1.65.0, removes pre-commit package, uses same system support as [cardano-node](https://github.com/input-output-hk/cardano-node), removes default package for monorepo.


cameron1024 (approved):
LGTM 🦀 no more pre-commit 😁

PR #62, opened 2022-11-09
NPG-3846 Add OpenTelemetry layer to jormungandr

FelipeRosa:


saibatizoku (approved):
Nice refactoring with LogGuard, cleans up a lot around log settings and format as well.

PR #63, opened 2022-11-09
Update top-level script to build/test/install everything in the monorepo

saibatizoku:
Took 1 hr 17 mins to run on my local box, but cargo doesn't complain (except for a temporary glitch in the vitup code).


Mr-Leshiy (approved):
LGTM

PR #64, opened 2022-11-09
NPG-3847 Add OpenTelemetry layer to vit-ss server

FelipeRosa:


saibatizoku (approved):
Nice cleanup. My only comment is, I don't know if it goes on this PR, that we should probably set `tracing` and `opentelemetry-*` crates as workspace dependencies. Doesn't make sense to have different versions throughout the crates we add this layer to.

FelipeRosa:
I believe I can do that in a separate PR as it would touch multiple crates. WDYT?

PR #67, opened 2022-11-10
Reverts manual function name generation 

dkijania:
Use function_name crate to generate automatically names for logger in vitup


minikin:
@cameron1024

cameron1024:
I removed this on purpose, just forgot to check the test code for compile errors. I removed it for a couple of reasons:
 - it was an easy target when removing dependencies to help with build times (we currently have 1000, which is kinda absurd)
 - IMO, this macro is a bit of a footgun, since it turns the function name into an implementation detail. If you're not familiar with the `#[named]` macro, you might think that changing the name of a function couldn't possibly have an effect on the behaviour, but now it does

Don't feel super strongly about adding it back, but on the CI branch I've just gone and replaced it with just manually typing out the function names

dkijania:
Yes. that makes sense. removing PR

PR #69, opened 2022-11-10
Embed vit-ss migration files

FelipeRosa:
vit-ss used to embed the migration files but after [these changes](https://github.com/input-output-hk/catalyst-core/commit/586ae82961c79894f251d148a5a7cbd04c31b21e#diff-e125a6865c74aefe3235510a1fbb72cded9672ac918e9be8214466a53a0ac1c1L3) migration files were being read from relative directories which broke things.

This makes the vit-ss CLI init db command work correctly again and also fixes the vit-ss integration tests (still testing only SQLite).

*PS: Since diesel migrations lib seems to only provide functions for running migrations that receive directory paths as arguments, I decided to embed the migration files and write them to a temporary directory when running the migrations. Not sure if this is the best solution.*

### DIFF
--- a/src/vit-servicing-station/vit-servicing-station-lib/src/db/migrations.rs
+++ b/src/vit-servicing-station/vit-servicing-station-lib/src/db/migrations.rs
@@ -1,20 +1,46 @@
+use tempfile::TempDir;
+
 use crate::db::DbConnection;
+
+const PG_INITIAL_SETUP_UP: &str =
+    include_str!("../../migrations/postgres/00000000000000_diesel_initial_setup/up.sql");
+const PG_SETUP_UP: &str =
+    include_str!("../../migrations/postgres/2022-11-03-210841_setup_db/up.sql");
+const SQLITE_SETUP_UP: &str =
+    include_str!("../../migrations/sqlite/2020-05-22-112032_setup_db/up.sql");
 
 pub fn initialize_db_with_migration(db_conn: &DbConnection) {
     // Discard all output
     let mut sink = std::io::sink();
 
     match db_conn {
-        DbConnection::Sqlite(db_conn) => diesel_migrations::run_pending_migrations_in_directory(
-            db_conn,
-            &std::path::PathBuf::from("./migrations/sqlite"),
-            &mut sink,
-        ),
-        DbConnection::Postgres(db_conn) => diesel_migrations::run_pending_migrations_in_directory(
-            db_conn,
-            &std::path::PathBuf::from("./migrations/postgres"),
-            &mut sink,
-        ),
+        DbConnection::Sqlite(db_conn) => {
+            let t = TempDir::new().unwrap();
+
+            write_tmp_migration(&t, "2020-05-22-112032_setup_db", SQLITE_SETUP_UP);
+
+            diesel_migrations::run_pending_migrations_in_directory(db_conn, t.path(), &mut sink)
+        }
+        DbConnection::Postgres(db_conn) => {
+            let t = TempDir::new().unwrap();
+
+            write_tmp_migration(
+                &t,
+                "00000000000000_diesel_initial_setup",
+                PG_INITIAL_SETUP_UP,
+            );
+            write_tmp_migration(&t, "2022-11-03-210841_setup_db", PG_SETUP_UP);
+
+            diesel_migrations::run_pending_migrations_in_directory(db_conn, t.path(), &mut sink)
+        }
     }
     .unwrap();
+}
+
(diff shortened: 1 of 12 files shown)


src/vit-servicing-station/vit-servicing-station-lib/src/db/migrations.rs
```diff
@@ -1,20 +1,46 @@
...
-            db_conn,
-            &std::path::PathBuf::from("./migrations/sqlite"),
-            &mut sink,
-        ),
-        DbConnection::Postgres(db_conn) => diesel_migrations::run_pending_migrations_in_directory(
-            db_conn,
-            &std::path::PathBuf::from("./migrations/postgres"),
-            &mut sink,
-        ),
+        DbConnection::Sqlite(db_conn) => {
```
dkijania, line 17: both match arms are really similar. I think they are only different in migration name and content. Maybe we can make it less duplicated?

src/vit-servicing-station/vit-servicing-station-lib/src/db/migrations.rs
```diff
@@ -1,20 +1,46 @@
...
+            let t = TempDir::new().unwrap();
+
+            write_tmp_migration(&t, "2020-05-22-112032_setup_db", SQLITE_SETUP_UP);
+
+            diesel_migrations::run_pending_migrations_in_directory(db_conn, t.path(), &mut sink)
+        }
+        DbConnection::Postgres(db_conn) => {
+            let t = TempDir::new().unwrap();
```
dkijania, line 25: Maybe we should handle this? instead of unwrap?

src/vit-servicing-station/vit-servicing-station-lib/src/db/migrations.rs
```diff
@@ -1,20 +1,46 @@
...
-        ),
-        DbConnection::Postgres(db_conn) => diesel_migrations::run_pending_migrations_in_directory(
-            db_conn,
-            &std::path::PathBuf::from("./migrations/postgres"),
-            &mut sink,
-        ),
+        DbConnection::Sqlite(db_conn) => {
+            let t = TempDir::new().unwrap();
+
+            write_tmp_migration(&t, "2020-05-22-112032_setup_db", SQLITE_SETUP_UP);
```
cong-or, line 20: what does this represent? _"2020-05-22-112032_setup_db"_

does it refer to something on your local machine?

src/vit-servicing-station/vit-servicing-station-lib/src/db/migrations.rs
FelipeRosa: Agreed

src/vit-servicing-station/vit-servicing-station-lib/src/db/migrations.rs
FelipeRosa: That's the a directory containing SQL db migration. It's created by Diesel when you generate a new migration.

FelipeRosa: PS: https://github.com/input-output-hk/catalyst-core/tree/main/src/vit-servicing-station/vit-servicing-station-lib/migrations

cong-or: 👍

src/vit-servicing-station/vit-servicing-station-lib/src/db/migrations.rs
cameron1024: I'm not sure we need to use `Result` here.

Presumably this function is being called only at server startup? If so, panicking is arguably the correct behaviour. 

Don't think it matters much, but I like to reserve `Result` for "recoverable errors", and this sort of thing isn't really recoverable

saibatizoku (approved):

src/vit-servicing-station/vit-servicing-station-lib/src/db/migrations.rs
dkijania: agreed, but at least some more meaningful error message instead of unwrap?

FelipeRosa: This is used by tests and by `vit-servicing-station-cli db init`. I like the idea of making it return result and leaving the responsability of panicking to the caller because it isn't necessarily unrecoverable in every context.

I had already started working on making these changes and made a new commit so you can see how it changes.

saibatizoku: +1

If we need further information, it should probably go to the logs/tracing.

PR #79, opened 2022-11-16
Create CODE_OF_CONDUCT.md

minikin:
Add CODE_OF_CONDUCT


cameron1024 (approved):
Maybe worth adding a link to this in the readme, and saying it's copied verbatim from the Rust project?

minikin:
@cameron1024 it's pretty standard, so I don't think that we need to mention Rust project here .

PR #83, opened 2022-11-18
Dont merge - Ci no fail fast

cameron1024:
Just creating this to run CI without failfast to collect all failures in one place


minikin:
@cameron1024 Should we close this one?

PR #85, opened 2022-11-18
Setup CI/CD for wallet Javascript bindings 

Mr-Leshiy:


.github/workflows/js_publish.yml
```diff
@@ -0,0 +1,79 @@
+name: Javascript wallet
```
minikin, line 1: I'd suggest having an explicit name for the  job, something like **"Release JavaScript Wallet"** and a workflow file name like `publish_js_wallet.yml`

.github/workflows/js_publish.yml
```diff
@@ -0,0 +1,79 @@
...
+
+on:
+  push:
+    branches: [ "main" ]
```
minikin, line 5: Do we really want to trigger this workflow on each PR merge?

.github/workflows/js_tests.yml
```diff
@@ -0,0 +1,35 @@
+name: Javascript wallet
```
minikin, line 1: I'd also suggest having an explicit name for the job, something like "JavaScript Wallet Tests." and the same for the workflow file name.

.github/workflows/js_tests.yml
```diff
@@ -0,0 +1,35 @@
...
+    steps:
+    - uses: actions/checkout@v3
+
+    - uses: actions-rs/toolchain@v1
+      with:
+        toolchain: stable
+
+    - uses: jetli/wasm-pack-action@v0.3.0
+      with:
+        # Optional version of wasm-pack to install(eg. 'v0.9.1', 'latest')
```
minikin, line 22: Can we remove this comment? It doesn't add any value here, IMO.

minikin (approved):
Approved! I left a few minor comments.

.github/workflows/js_publish.yml
Mr-Leshiy: It will only work after we have establish `tag`

PR #87, opened 2022-11-19
Implement tracing in vitup instead of printlns

dkijania:


src/vit-testing/vitup/src/builders/helpers/qr.rs
```diff
@@ -25,9 +26,17 @@ pub fn generate_qr_and_hashes<P: AsRef<Path>>(
...
+    let span = span!(Level::TRACE, "qr code generation");
+    let _enter = span.enter();
+
     let total = wallets.len();
     let folder = folder.as_ref();
 
+    info!(
+        "generating {} valid qr codes and payloads...",
+        wallets.len()
+    );
```
FelipeRosa, line 38: I don't know if this is useful here or somewhere else so I'm just sharing this: you can include fields in the span and events as described here https://docs.rs/tracing/latest/tracing/#recording-fields.

FelipeRosa (approved):
:heart:

src/vit-testing/vitup/src/builders/helpers/qr.rs
dkijania: Good point @FelipeRosa . Moving it to span

PR #88, opened 2022-11-20
Gateway node 3887

cong-or:
As we gradually start to allow full public node participation in the network, we will start by deploying gateway nodes as an entry point. Gateway nodes can only consume from the internal network, they cannot _propagate/gossip/publish_.

This is a _temporary_ solution to prevent flooding attacks etc.

Below is a _crude_ diagram to illustrate the point.

A gateway node that is conditionally compiled (**_d_** in the diagram), 

**d** will be open to the wild - **A, B, C** will not be. 

Screenshot 2022-11-15 at 13 52 08

### DIFF
--- a/src/jormungandr/testing/jormungandr-integration-tests/src/networking/p2p/connections.rs
+++ b/src/jormungandr/testing/jormungandr-integration-tests/src/networking/p2p/connections.rs
@@ -1,10 +1,4 @@
-use std::path::PathBuf;
-
-use crate::{jormungandr::grpc::setup, networking::utils};
-use chain_core::property::{BlockId, HasHeader};
-
-use chain_impl_mockchain::testing::{GenesisPraosBlockBuilder, StakePoolBuilder};
-use chain_time::{Epoch, TimeEra};
+use crate::networking::utils;
 use hersir::{
     builder::{NetworkBuilder, Node, Topology},
     config::{BlockchainConfiguration, SpawnParams, WalletTemplateBuilder},
@@ -22,7 +16,6 @@ const LEADER3: &str = "LEADER3";
 const LEADER4: &str = "LEADER4";
 
 const CLIENT: &str = "CLIENT";
-const CLIENT_B: &str = "CLIENT_B";
 const SERVER: &str = "SERVER";
 const SERVER_1: &str = "SERVER_1";
 const SERVER_2: &str = "SERVER_2";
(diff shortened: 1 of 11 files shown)


cong-or:
@dkijania do these tests make sense?

src/jormungandr/testing/jormungandr-integration-tests/src/networking/p2p/connections.rs
```diff
@@ -108,6 +115,91 @@ pub fn node_trust_itself() {
         .unwrap();
 }
 
+#[ignore]
+#[test]
+fn gateway_gossip() {
```
dkijania, line 120: hm, i don't think `cross_version` module is not a good place for this test method. Maybe it should be put into more meaningful module? like `gateway`?

src/jormungandr/testing/jormungandr-integration-tests/src/networking/p2p/connections.rs
```diff
@@ -108,6 +115,91 @@ pub fn node_trust_itself() {
...
+            SpawnParams::new(SERVER)
+                .gossip_interval(Duration::new(SERVER_GOSSIP_INTERVAL_SECS, 0))
+                .log_level(LogLevel::TRACE),
+        )
+        .unwrap();
+
+    //
+    // spin up gateway node
+    //
+    let mut gateway_node_binary = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
```
dkijania, line 158: i wonder if `gateway` app is not available on path when running this test?

src/jormungandr/testing/jormungandr-integration-tests/src/networking/p2p/stats.rs
```diff
@@ -159,6 +163,66 @@ macro_rules! build_network {
...
+        100.into(),
+    ) {
+        Ok(_) => panic!("gateway node should not be able to send fragments"),
+        Err(err) => assert_eq!(
+            err.to_string(),
+            "Too many attempts failed (1) while trying to send fragment to node: ".to_string()
+        ),
+    };
+}
+
```
dkijania, line 225: i would add one more test just to be sure that gateway node function correctly with trusted nodes. For example we are configure gateway node not to publish but we need to make sure that it still sync with rest of the network correctly

dkijania:
@cong-or looking good! i see one more potential test

PR #95, opened 2022-11-23
Fix vit-testing integration tests

Mr-Leshiy:
During fix of the `transactions_are_send_between_nodes_with_correct_order` test I have figure out that it is missed preinstalled following tools:
- jormungandr
- vit-servicing-station-server
- valgrind
- iapyx
- vitup

Seems that it will also help for other integration tests


minikin:
@Mr-Leshiy Should we close this PR?

PR #97, opened 2022-11-23
build(vit-servicing-station-server): adds missing psql dependency to package

jmgilman:
At some point, `vit-servicing-station-server` inherited a dependency on `libpq`.

As a side note, we should probably be building the Nix packages in CI here so it's obvious when a change breaks the existing Nix package builds. This seems to have been broken for awhile.


jmgilman:
Master has broken tests?

minikin:
@jmgilman 
>Main has broken tests?

Unfortunately. We're working on fixing them.

minikin (approved):
LGTM

PR #106, opened 2022-11-28
Remove not used services in vit-testing

dkijania:
Removing two services which operates on haskell runtime dependencies for registrations. Since now, we have our rust implementation there are not needed.

Moved CardanoCli wrapper to mainnet tools as they might be useful in nearest future


minikin:
@kukkok3 @nicopado Can you have a look, please&

jmgilman:
Can we remove the respective nix entries from `nix/vit-testing/packages.nix`?

nicopado (approved):

minikin:
@dkijania Can we fix conflicts, 🙏?

PR #111, opened 2022-11-30
build: adds std variants of packages

jmgilman:
Introduces `std` and refactors packaging to be more discoverable/maintainable. 

- [x] Devshell has been converted
- [x] All existing packages are done
- [x] Operables have been created for all services
- [x] Containers have been created for all services


jmgilman:
Something to consider here is that we're not running tests when building packages. If we want, it's possible to enable tests to be run with a trivial change (one line of code). However, it would enable tests for _all_ packages. 

The biggest question that needs to be answered is if any of the cargo tests would fail when run inside a sandbox (where Nix runs them). A sandbox with no access to the network, environment, etc. Typically this is fine for unit tests, but if we have more complex tests baked in then the package builds will likely start failing. 

Thoughts?

jmgilman:
In the meantime, this PR is ready for review. I'd like to get it merged soon so I can quickly shift to getting the packages/containers built in the CI.

minikin:
@jmgilman 
That sounds like a great idea. We still need to fix a few failing test suits. I'll add a ticket in Jira to do not lost this idea.

minikin:
@const-iohk @dkijania @kreisys Can you have a look, please?

nix/vit-testing/operables.nix
```diff
@@ -0,0 +1,57 @@
...
+      };
+    };
+in {
+  registration-service = mkSimpleOperable {
+    name = "registration-service";
+    runtimeInputs = [
+      catalyst-toolbox
+      jcli
+      cardano-cli
+      voting-tools-rs
```
dkijania, line 33: registration service doesn't need this dependency

nix/vit-testing/operables.nix
```diff
@@ -0,0 +1,57 @@
...
+      jcli
+    ];
+    args = {
+      "--config" = "/secrets/registration-verify-service.config";
+    };
+  };
+  snapshot-trigger-service = mkSimpleOperable {
+    name = "snapshot-trigger-service";
+    runtimeInputs = [
+      jcli
```
dkijania, line 51: here we need a `voting-tools` and `voting-tools-rs` dependency

nix/vit-testing/operables.nix
```diff
@@ -0,0 +1,57 @@
...
+      cardano-cli
+      voting-tools-rs
+    ];
+    args = {
+      "--config" = "/secrets/registration-service.config";
+    };
+  };
+  registration-verify-service = mkSimpleOperable {
+    name = "registration-verify-service";
```
dkijania, line 40: I know that you are replicating all services we previously had, but lately registration service and registration-verify-service are deprecated, so i think we don't need those jobs

dkijania (commented):
We also need to deploy voting-tools:

https://github.com/input-output-hk/voting-tools

 and it should be available for snapshot trigger service

nix/vit-testing/operables.nix
jmgilman: This is good feedback. Now is our chance to update these things, thanks! :)

nix/vit-testing/operables.nix
jmgilman: Added. I can't get `voting-tools` building at the moment, it's failing on some obscure error about missing licenses, but I've forwarded it internally to see if I can figure out what's going on.

minikin (approved):

PR #112, opened 2022-11-30
fix doc build issues

cameron1024:
Docs now build without errors with `RUSTDOCFLAGS=-Dwarnings cargo doc`

If we want, we can add a check for this to CI


rust-toolchain
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.65"
+channel = "nightly"
```
minikin, line 2: We agreed to use the latest stable version. Using nightly will increase inconsistency in the project significantly, and I'm pretty sure it will slow down development.

cameron1024: Ah this was a mistake, we need to build docs with nightly (many crates require nightly to build their docs, since this is what docs.rs does, so the build just fails on stable).

I changed this because I'm using the rustc provided by the devshell, which reads this file. I'll change it back, but if we want to run it in CI, we'll need nightly. However, if we pin to a specific nightly version, we'll avoid the issues with random breakage (since nix will pin it to a specific nightly version)

minikin: I'd suggest fixing all test suits first, having stable builds, and then thinking about what to do with docs.
It's better to have consistency and production-ready product and docs with errors than to go to the path where we need to repeatedly re-do / re-fix some things.

Ps: tests are failing again now.

minikin: We discussed this offline. We'll bring up this topic for the following tech discussion.

stevenj:
LGTM.  Other than the merge conflicts with main,  any reason we are holding off on the merge?

cameron1024:
Nope, this one can merge

minikin (approved):

PR #113, opened 2022-11-30
include jormungandr tests in CI

kukkok3:


kukkok3:
Runned the jormungandr twice in CI and locally with no failings, so I think it is safe to add to the CI

minikin (approved):
LGTM!

PR #115, opened 2022-11-30
Adding jormungandr integration tests to CI

kukkok3:


minikin:
@kukkok3 Do we still need this PR?

PR #124, opened 2022-12-02
Voting tools signature verification adjustment.

dkijania:
Included voting purpose in  verification process. 
 
 After integration tests we discovered that correct signature verification is incorrect. Adding voting purpose to signature verification process fixed the problem


cameron1024:
Ahh, this was next on my list to check 😁 good spot. I'll review this as soon as I'm back in front of my PC

dkijania:
Apologies for stepping out and taking you prey :( . This bug was blocking integration tests with GVC, therefore we need a rapid dev and fix

Mr-Leshiy (approved):

cameron1024 (approved):
LGTM - no problem, thanks for the fix 😁

PR #125, opened 2022-12-02
Wallet Javascript bindings, singVotes function

Mr-Leshiy:
- added a separate JS package with the convenient wrapper over pure JS Wasm code
- implement singVotes function as a part of CIP-62


Mr-Leshiy:
@stevenj pls can you take a look closely into the test flow, does it fulfill how the interface should look like and should be used?

https://github.com/input-output-hk/catalyst-core/blob/1c27a80f09cc9626d2aca0b4419742ecc4cd2b4f/src/chain-wallet-libs/bindings/wallet-js/js-test/test/vote_cast.js#L38

dkijania (approved):

PR #129, opened 2022-12-05
Update Javascript CD process

Mr-Leshiy:
Add a new Javascript wallet API package publishing.
Update chain-wallet-libs versions.


src/chain-wallet-libs/bindings/wallet-js/js/package.json
```diff
@@ -1,11 +1,14 @@
 {
-  "name": "js",
-  "version": "0.0.1",
+  "name": "wallet-js-api",
```
minikin, line 2: Can we consider using `wallet-js` as the package name? It looks like the suffix `-api` doesn't add value.

Mr-Leshiy: we have the same name for the wasm package, should we rename it for `wallet-wasm-js` ?

minikin: We probably also can consider removing `-js` because both packages are JS packages: `wallet` and `wallet-wasm`

minikin (approved):

PR #140, opened 2022-12-08
fix hex-encoded string conversion for chain_proposal_id

saibatizoku:
An issue for extracting the correct bytes from a hex-encoded string is resolved.

The fix was added to convert the 64-character hex-encoded digest of a Blake2b256 into the underlying 32-byte array used to define the `chain_proposal_id` field of the `FullProposalInfo` type.

### DIFF
--- a/src/catalyst-toolbox/catalyst-toolbox/src/rewards/dreps.rs
+++ b/src/catalyst-toolbox/catalyst-toolbox/src/rewards/dreps.rs
@@ -167,11 +167,8 @@ mod tests {
         let voters = snapshot.to_full_snapshot_info();
 
         let snapshot = ArbitrarySnapshotGenerator::default().snapshot();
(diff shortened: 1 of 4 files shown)


minikin:
@saibatizoku It looks like we need to update the tests 🧪

saibatizoku:
> @saibatizoku It looks like we need to update the tests test_tube

@minikin Local tests are passing, I'm investigating the CI tests.

saibatizoku:
@minikin The tests are fixed and the CI is green across the board.

stevenj (approved):
This LGTM.

In a separate PR, Can you please add to the docs page I started to capture technical details of the votes? Describe what the `chain_proposal_id` is,  why its different from the idea scale proposal id (if you know) and how its derived/generated, and anything else worth mentioning about it.

PR #143, opened 2022-12-09
add information about chain_proposal_id field

saibatizoku:
Things I've found out from my recent task related to `chain_proposal_id` .


stevenj (approved):
Thanks

PR #146, opened 2022-12-12
Update Javascript wallet bindings, replace wallet approach with the basic transaction building

Mr-Leshiy:
- make a singVotes as a free function, not a method of the Wallet type
- remove Wallet type and Wallet usage from the Javascript bindings
- implement a simple TransactionBuilder type with the only transaction building functionality.
- refactoring


stevenj (changes requested):
The wallet will get a vote structure from the dApp which has the `valid_until` and the `spending counter` embedded within it.

See CIP-62,  but its defined currently as:

```ts
interface BlockDate {
  epoch: number
  slot: number
}
```

* epoch - an epoch value on the voting block chain.
* slot - a block slot number on the voting block chain.

#### Proposal

```ts
interface Proposal {
  votePlanId: string
  voteOptions: number[]
  votePublic: boolean
  proposalIndex: number
}
```

Proposal information.

* votePlanId - Hex encoded string to represent vote plan unique identifier. This will be the same as anchored on voting chain, and is managed by the dApp.
* voteOptions - List of possible "choices" which can be made for this proposal.
* votePublic - Boolean indicating whether the vote is public or private.

### Vote

```ts
interface Vote {
  proposal: Proposal;
  choice: number;
  expiration: BlockDate;
  purpose: VotingPurpose;
  spendingCounter?: number;
  spendingCounterLane?: number;
}
```

Can the `valid_until` and spending counters be used from this vote structure, or does this structure need to change?

Mr-Leshiy:
@stevenj thanks for your comments.
`BlockDate` easy fix, will do it.
Will try to update `Vote` and `Proposal` types as in the CIP-62 described.

Mr-Leshiy:
@stevenj updated, you can take a look

stevenj (approved):
LGTM

PR #151, opened 2022-12-12
Implement dbsync mock with cardano node mock

dkijania:
Implement necessary mocks which can simulate communication flow between db sync and cardano node. Removed JsonBasedDbSync as it was redundant to in memory db sync and just move persist and restore methods to in memory db sync.

Introduced  also basic multi-threaded mechanism  of cardano node leadership process as well as db sync update thread. Thanks for that we can use background tokio tasks for producing new blocks and in the same time updates db sync


src/vit-testing/mainnet-lib/src/cardano_node/ledger.rs
```diff
@@ -0,0 +1,58 @@
...
+    pub fn push_transaction(&mut self, transaction: Transaction) {
+        self.mempool.push(transaction);
+    }
+
+    /// Mint new block
+    pub fn mint_block(&mut self) -> Block {
+        let last_block = self
+            .blocks
+            .last()
+            .expect("internal error no block0 in blockchain");
```
kukkok3, line 52: Wouldnt be better to return a proper Error here?

src/vit-testing/mainnet-lib/src/cardano_node/settings.rs
```diff
@@ -0,0 +1,26 @@
...
+    pub slots_per_epoch: u32,
+}
+
+impl Default for Settings {
+    fn default() -> Self {
+        Self {
+            block0_hash: TestGen::hash().to_string(),
+            block0_time: SystemTime::now(),
+            slot_duration: 1,
+            slots_per_epoch: 43200,
```
kukkok3, line 23: What about using a const here like `const CARDANO_MAINNET_SLOTS_PER_EPOCH: u64 = 43200;`

src/vit-testing/mainnet-lib/src/db_sync/in_memory.rs
```diff
@@ -0,0 +1,353 @@
...
+                        .iter()
+                        .find(|x| x.header().header_body().block_number() == *block)
+                        .expect("data inconsitency. cannot find block with block no from tx");
+
+                    return Some((block, tx));
+                }
+                None
+            })
+            .collect()
+    }
```
kukkok3, line 154: I might be wrong but how can you return `Some()` and `None ` if the return type is not an `Option`?
Also I would expect to have a return type with one transaction (if it exist) and one or many blocks, maybe something like `Some(HashMap<Transaction, Vec<Block>>)` ?
Also I would avoid abbreviations in the `.expect("Data inconsistency. Cannot find block with block number from the transaction") `

kukkok3 (changes requested):
Looks good, just some comments to address

src/vit-testing/mainnet-lib/src/cardano_node/settings.rs
dkijania: good catch!

src/vit-testing/mainnet-lib/src/db_sync/in_memory.rs
dkijania: `filter_map` is doing magic with mapping elements and in the same time filtering it. That's why we need to tell them if we want to filter out some element (by returning None) or how we want to map elements which are not filtered out (by returning Some(....)). During `collect` phase of iterator, it removes Option<...> wrapper so we don't need to do it on our own. I will remove this nasty expect. thanks!

kukkok3 (approved):

PR #152, opened 2022-12-13
Remove unused bip39 crate

minikin:
- Remove unused crate bip39


stevenj:
Looks like a good idea.  Even if we ever needed it there are plenty of crates which implement it we should use by preference.

cong-or (approved):
👍

PR #154, opened 2022-12-13
Npg 3427 voting key saturation endpoint tests

nicopado:
Added test for key saturation endpoint, updated VotingPower struct to sync it to the current production one


kukkok3 (approved):
LGTM but there are some tests to be fixed in vit-servicing-station

PR #156, opened 2022-12-13
Improve node config builder

kukkok3:
Introducing a new function `default_with_config` for `JormungandrBootstrapper ` to use instead of `JormungandrBootstrapper::default().with_node_config(node_config)` 
This is because the `Default` trait for `JormungandrBootstrapper` calls `NodeConfigBuilder` creating a `NodeConfig` that it will be overwritten by the `NodeConfig` passed by `with_node_config()`
`NodeConfig` will call the function `get_available_port()` that makes operations on an `Atomic `variable and "flags" some ports as being used, both operations that would be better to not duplicate for performance and debbugging reasons


src/jormungandr/testing/jormungandr-automation/src/jormungandr/starter/params/builder.rs
```diff
@@ -45,6 +45,21 @@ impl Default for JormungandrBootstrapper {
 }
 
 impl JormungandrBootstrapper {
+    pub fn default_with_config(node_config: NodeConfig) -> Self {
```
dkijania, line 48: I would like to have a documentatio part why it exist. Basically what you wrote on PR is great so IMHO we should just pass this explanation here

dkijania (approved):

PR #158, opened 2022-12-13
fix: updates devshell with missing dependencies

jmgilman:


cameron1024 (approved):

cameron1024:
just noting: this also removes jemalloc, since it's only used for an example in imhamt, which doesn't seem important

PR #159, opened 2022-12-13
add socket addr to Fragment Origin

cong-or:
PR facilitates tracing back the origin of a fragment to a given network node by adding a socket addr to the FragmentOrigin type.


src/jormungandr/jormungandr-lib/src/interfaces/fragment_log.rs
```diff
@@ -21,6 +21,14 @@ pub enum FragmentOrigin {
     JRpc,
 }
 
+impl FragmentOrigin {
+    pub fn default_origin_addr() -> Self {
+        FragmentOrigin::Network {
+            addr: SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 27000),
```
stevenj, line 27: You could use https://github.com/EstebanBorai/local-ip-address to automatically get the default src address of the running machine.  Then it would not need to be a config parameter, the parameter could simply do used to over-ride this auto detected value in the rare circumstance its wrong.

cong-or: did you mean, ignore the config and just use the default lookup?

stevenj: Yes,  exactly.  The only other comment is that when we use the local ip, I wouldn't trace with error level as its perfectly acceptable behavior.  Warning or even just info would be sufficient so its clear in the logs we used the local lookup address and not one from config.  This helps find issues when config isn't working like you hoped.

stevenj (approved):

PR #165, opened 2022-12-14
use nightly in devshell

cameron1024:
following on from technical discussion, move to using nightly in the devshell, but keep stable for builds + CI

when running `nix flake update`, it will update to the latest nightly that has miri


jmgilman:
Minor nit, can you format the changes with `treefmt`? It should already be configured with the correct formatter for nix—otherwise, LGTM.

nix/lib/toolchains.nix
```diff
@@ -1,17 +1,21 @@
-{
```
jmgilman, line 1: Formatting changes here is likely due to using a different formatter from the one `treefmt` is using.

jmgilman (changes requested):

nix/lib/toolchains.nix
cameron1024: I think this is using the one rnix-lsp provides, I'll run treefmt 👍

jmgilman (approved):

PR #167, opened 2022-12-14
build: adds Nix CI

jmgilman:


.github/workflows/nix.yml
```diff
@@ -0,0 +1,105 @@
...
+      - 'Cargo.*'
+      - 'flake.*'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.sha }}
+  cancel-in-progress: true
```
nrdxp, line 27: ran into issues with this in Mamba. You may want to consider making jobs wait instead of canceling, at least on the main branch, to ensure each commit is cached. You can look here to see what I am doing currently:
https://github.com/input-output-hk/mamba-world/blob/a6e47b05dcb389051a8ce213a3e86270810c551e/.github/workflows/std.yml

nrdxp (approved):
Small comment based on feedback from Mamba devs but overall looks good.

minikin:
@cameron1024 Can you review this, please?

cameron1024 (approved):

PR #172, opened 2022-12-16
do not install pre-commit hooks

stevenj:
Pre commit hooks get installed without warning when running a nix dev shell.  This has the effect of changing the development environment outside the dev shell.

The pre-commit hooks do not take into consideration the local developers environment or tooling, such as IDE of choice.  Also, the pre-commit hooks assume tooling which may only be present inside the dev shell.  This can lead to an inability to commit and interferes with local development.

This patch removes that, so that running the dev shell does not change the development environment outside the dev shell.

### DIFF
--- a/nix/automation/configs.nix
+++ b/nix/automation/configs.nix
@@ -41,13 +41,13 @@ in {
       #     };
       #   };
       # };
-      pre-commit = {
-        commands = {
-          treefmt = {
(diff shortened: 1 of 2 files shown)


stevenj:
> Just remove this line instead: https://github.com/input-output-hk/catalyst-core/blob/main/nix/automation/devshells.nix#L22

Done

minikin (approved):

jmgilman (approved):
LGTM

PR #173, opened 2022-12-16
build: reintroduces namespacing for jormungandr/vitss

jmgilman:


const-iohk (approved):
Yes, as per our previous discussion this will initially reduce combinations, lgtm, approving.

PR #175, opened 2022-12-19
build: generates vit-ss config file in container

jmgilman:


const-iohk (approved):
Nice work in generating the config file locally, can you extract service version not to be hardcoded and read from the env var instead? Otherwise all OK, and approved.

PR #176, opened 2022-12-19
build: only publish images on push to main

jmgilman:


jmgilman:
FYI I snuck in [8a233b8](https://github.com/input-output-hk/catalyst-core/pull/176/commits/8a233b80f106d480f1f19accd3fb8f1c0dad8feb) since at some point someone managed to commit these files (which are included in `.gitignore`).

const-iohk (approved):
LGTM, resolves the issue with gh sha references.

PR #180, opened 2022-12-20
build: modifies vit-ss to use env variables for db/service version

jmgilman:


jmgilman:
The vit-ss CLI gives the following information about the database url:

```
Database url [env: DATABASE_URL=]  [default: ./db/database.sqlite3]
```

It might be worth testing setting `DATABASE_URL` directly in the container environment to see if it works. This would be more efficient than passing the flag, IMO.

const-iohk (approved):
Yes, service version and db url are preferable this way, to test container creation. Thanks!

PR #189, opened 2022-12-22
Update wallet-js CD job

Mr-Leshiy:
# Description

Simple cleanup of the wallet-js npm package publishing process

### DIFF
--- a/.github/workflows/js_wallet_publish.yml
+++ b/.github/workflows/js_wallet_publish.yml
@@ -4,41 +4,48 @@ on:
   push:
     tags:
       - walletjs-*
+  pull_request:
+    branches:
+      - main
 
 jobs:
-  initial_release:
-    name: Publish wasm package
-    runs-on: ubuntu-latest
-    outputs:
-      upload_url: ${{ steps.create_release.outputs.upload_url }}
-      version: ${{ steps.get_version.outputs.version }}
-    steps:
-      - name: Checkout code
-        uses: actions/checkout@v3
+  # initial_release:
+  #   name: Publish wasm package
+  #   runs-on: ubuntu-latest
+  #   outputs:
+  #     upload_url: ${{ steps.create_release.outputs.upload_url }}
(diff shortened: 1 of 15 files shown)


.github/workflows/js_wallet_publish.yml
```diff
@@ -6,39 +6,43 @@ on:
...
-  initial_release:
-    name: Publish wasm package
-    runs-on: ubuntu-latest
-    outputs:
-      upload_url: ${{ steps.create_release.outputs.upload_url }}
-      version: ${{ steps.get_version.outputs.version }}
-    steps:
-      - name: Checkout code
-        uses: actions/checkout@v3
+  # initial_release:
```
minikin, line 9: What was the motivation to comment out this job?

.github/workflows/js_wallet_publish.yml
```diff
@@ -6,39 +6,43 @@ on:
...
+    name: Release wallet js assets
+    # needs: initial_release
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        target: [nodejs, web, bundler]
+        target: [
+                nodejs, 
+                # web,
```
minikin, line 43: Should we uncomment `web` and `bundler`?

minikin (changes requested):

.github/workflows/js_wallet_publish.yml
Mr-Leshiy: I have removed it for testing purposes, forgotten to uncomment

.github/workflows/js_wallet_publish.yml
Mr-Leshiy: At this moment it is not possible to publish a few versions of packages under the same rpm package registry (at least at that way which we are doing at this moment), so only one job can finish successfully. I would let at this moment only `nodejs` version and figure out how it is better to publish a few later.

minikin (approved):

PR #192, opened 2022-12-28
Add rust style guide

minikin:
The initial proposal for Rust Style Guide for Catalyst Core.

Let's have a productive and constructive discussion around this.

**Feel free to contribute to the style guide.**

- [x]  Team agreed on the initial proposal.


book/src/rust/rust_style_guide.md
```diff
@@ -0,0 +1,660 @@
...
+}
+```
+over
+
+```rust
+fn use_mutex(m: sync::mutex::Mutex<int>) {
+    do_work(m.lock());
+    // do other work
+}
+```
```
cameron1024, line 392: I don't think this really adds much. This made more sense pre-NLL, but IMO this should be left to the developer's judgement

book/src/rust/rust_style_guide.md
```diff
@@ -0,0 +1,660 @@
...
+s.iter().map(|x| x * 2)
+        .collect::<Vec<_>>()
+```
+
+Over
+
+```rust
+let v: Vec<_> = s.iter().map(|x| x * 2)
+                        .collect();
+```
```
cameron1024, line 435: I actually have the opposite preference, but I don't think one is *better* than the other. IMO we shouldn't have a rule for this.

Also it's not always possible to specify generics at the function call-site, so we'd still end up with inconsistencies

book/src/rust/rust_style_guide.md
```diff
@@ -0,0 +1,660 @@
...
+##### Prefer immutable bindings.
+
+Use mut bindings to signal the span during which a value is mutated:
+
+```rust
+let mut v = Vec::new();
+// push things onto v
+let v = v;
+// use v immutably henceforth
+```
```
cameron1024, line 450: This seems more confusing than helpful. If I saw `let v = v;`, I'd assume it was a mistake rather than intentionally removing `mut`

book/src/rust/rust_style_guide.md
```diff
@@ -0,0 +1,660 @@
...
+}
+```
+
+Notice the ordering of components:
+
+1. Associated Types
+
+2. Associated Constants
+
+3. Methods
```
cameron1024, line 490: Again with ordering, I think flexibility is more important here

book/src/rust/rust_style_guide.md
```diff
@@ -0,0 +1,660 @@
...
+
+Depending on the context and presentation, you can mix the ordering of required and optional methods. Also, notice the name formatting, although clippy should detect if naming differs from this pattern.
+
+##### Implementing Traits
```
cameron1024, line 494: I'm not sure how much this section adds

book/src/rust/rust_style_guide.md
```diff
@@ -0,0 +1,660 @@
...
+
+- `clippy::missing_panics_doc` - warns if your function might panic, and you haven’t included a “panics” section in the doc comment. However, it doesn’t work all the time, and IMO leads to a false sense of security. Absence of a warning seems to imply absence from panics, but it doesn’t. This isn’t true for the related missing_error_docs and missing_safety_docs , so those should remain enabled. If the API legitimately is intended to panic under some conditions, you should still add a panic section to the docs (e.g. maybe some cryptography library requires that a certain input number is prime, the docs should say panics if foo is not prime). The reason this lint is disabled is because the implementation is buggy, not because we don’t want panic docs. This also doesn’t apply to panics that would be considered programming errors if they ever happened
+
+#### Where Clauses
+Always use where clauses instead of inline trait constraints. So instead of
```
cameron1024, line 580: IMO this section is overly restrictive:
 - `<T: Clone>` is more readable in some contexts
 -  Ordering trait bounds alphabetically is also weird, sometimes there's a more natural order

cameron1024 (commented):
Hmm, this seems pretty restrictive. I think we should have a fairly minimal style guide, and only include things that aren't enforced by clippy/rustfmt.

For example, the sections on trailing commas, alignment, line wrapping, etc could all be deleted, since no PRs will pass CI if they don't do what rustfmt says. There's also a risk that the style guide gets out of sync with rustfmt, which leads to an annoying cycle of:
 - make it match style guide
 - push PR
 - CI fails because of rustfmt
 - run rustfmt, but that undoes style changes

The "official" style guide is also pretty old, and IMO hasn't evolved with the language. I'm not sure it's a good "starting point" for our own style guide.

For me, things that are important are:
 - what is our desired set of clippy lints (since these need to be added in source code, IIRC there isn't a way to have a project wide `clippy.toml` that specifies which lints to enable)
 - what sort of documentation we expect (I'd mention doctests, we don't use them much and I think we should)
 - other small bits and pieces (e.g. no magic numbers, preferring `&str` over `impl AsRef<str>`, etc)

I think it's really important that any style stuff is really lightweight, and that it doesn't get in people's way much. Tooling really helps with that, but it means that any style rule that can't be enforced by a tool needs a really compelling reason to exist

book/src/rust/rust_style_guide.md
```diff
@@ -0,0 +1,660 @@
...
+  - [Magic Numbers](#magic-numbers)
+
+### Dependencies
+
+Install clippy and rustfmt
+
+```sh
+rustup component add clippy
+rustup component add rustfmt
+```
```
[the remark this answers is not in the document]
minikin: I have a really mixed opinion about `nix` use for development. 
We can probably consider creating a separate doc about `nix` usage in this project. 
However, I don't think we need to force users to use `nix.`

book/src/rust/rust_style_guide.md
```diff
@@ -0,0 +1,660 @@
...
+
+##### Spaces
+
+- Use spaces around binary operators, including the equals sign in attributes:
+
+```rust
+#[deprecated = "Use `bar` instead."]
+fn foo(a: uint, b: uint) -> uint {
+    a + b
+}
```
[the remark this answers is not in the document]
minikin: I decided to keep it because engineers outside IOG  will also use this style guide (hopefully).

minikin:
@cameron1024 

Thank you for your comments! Let's collect initial feedback from the rest of the team and then have a dedicated technical discussion on the topic and address all comments.

>For me, things that are important are:
>
>- what is our desired set of clippy lints (since these need to be added in source code, IIRC there isn't a way to have a project wide clippy.toml that specifies which lints to enable)
>- what sort of documentation we expect (I'd mention doctests, we don't use them much and I think we should)
>- other small bits and pieces (e.g. no magic numbers, preferring &str over impl AsRef<str>, etc)

Please, feel free to add what is important for you to this PR.

book/src/rust/rust_style_guide.md
FelipeRosa: Why is the first one preferred? I usually find the second one less error-prone.

@cameron1024 What is "pre-NLL"?

minikin:
@cong-or 
>I have noticed inconsistencies with comments and doc comments within the code base.
> There are some suggestions not covered by clippy [here](https://github.com/rust-lang/fmt-rfcs/blob/master/guide/guide.md#comments)

Could you add those suggestions (just push to this pr) to our guide, please?

cong-or:
> @cong-or
> 
> > I have noticed inconsistencies with comments and doc comments within the code base.
> > There are some suggestions not covered by clippy [here](https://github.com/rust-lang/fmt-rfcs/blob/master/guide/guide.md#comments)
> 
> Could you add those suggestions (just push to this pr) to our guide, please?

yes will do

book/src/rust/rust_style_guide.md
```diff
@@ -0,0 +1,660 @@
...
+
+[profile....]
+...
+```
+Specifically, we have:
+
+1. Use double quotes instead of single quotes.
+2. Use the above as the standard ordering of the [package] map.
+3. [[bin]] before [features] before [dependencies] before [dev-dependencies] before [build-dependencies] before [profile] settings.
+4. Order features and dependencies alphabetically.
```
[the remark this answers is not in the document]
stevenj: I kind of agree,  but i would go further.  I think dependencies should have a comment stating WHY its being used.  My experience is as projects grow they gather dependencies like flies to a rotting carcass.  And no one is brave enough to remove them or look at them in detail because they don;t know why they are a dependency.  This also goes for features,  why do we use/need those features.

So,  building on what @cameron1024 said,  I would prefer:

```toml
[dependencies] 

# Local Crates
chain-addr = { path = "../../chain-libs/chain-addr" }  # Vote storage chain address type definition

# Tokio Async-IO Framework
#     Rest API endpoints are async.
tokio = { version = "^1.15", features = ["rt-multi-thread", "time", "sync", "rt", "signal", "test-util"] } # Why we use these features.
tokio-stream = { version = "0.1.4", features = ["sync"] }  # Feature "sync" is used because ...
tokio-util = { version = "0.6.0", features = ["time"] } # Feature "time" is used because ... 
```

So groups of related dependencies are clear and also why the choice what the program is using the crate for is clear.

book/src/rust/rust_style_guide.md
stevenj: I think the issue could be that in the second the drop is implicit, it could happen at the end of the use_mutex function,  not necessarily immediately after the lock is no longer needed.  Is it defined by the language when the drop will occur?  Dropping the guard explicitly ensures that the mutex is released as early as it can be.  Its also still safe because if the code was:

```rust
fn use_mutex(m: sync::mutex::Mutex<int>) -> Option<usize> {
    let guard = m.lock();
    do_work(guard)?;
    drop(guard); // unlock the lock
    // do other "Long running" work
    let result = ...

    Some(result)
}
```

The lock would get dropped automatically when do_work returned None, because it would immediately go out of scope.

book/src/rust/rust_style_guide.md
```diff
@@ -0,0 +1,697 @@
...
+```
+
+unless the conditions for initialization are too complex to fit into a simple conditional expression.
+
+##### Use type annotations for clarification; prefer explicit generics when inference fails.
+
+Prefer
+
+```rust
+s.iter().map(|x| x * 2)
```
stevenj, line 463:
```suggestion
let v = s.iter().map(|x| x * 2)
```

book/src/rust/rust_style_guide.md
stevenj: I prefer the second as well,  as its more "natural" to people coming from other languages.  I think the goal should be clarity to the human reading it (which is subjective), and in this case it depends too much on the actual code to say what would be "clearer".  But in this case, I think the second would be clearer to more people.

I would prefer if this rule was more about documentation, because these constructs quickly become difficult to reason through.  So, I would prefer the rule say,  "these kinds of constructs should also ALWAYS be accompanied by a comment explaining WHAT it is doing.

```rust
// v becomes a vector where every element of s is doubled.
let v: Vec<_> = s.iter().map(|x| x * 2)
                        .collect();
```

book/src/rust/rust_style_guide.md
stevenj: I disagree,  types and constants should be grouped together.   I don't see the need for them to be defined ad-hoc.  And if they are grouped,  the logical order is to define types first (because consts do not have any bearing on type definitions).  Then consts, because they could be consts of the types just defined,  and then methods, because they probably use the types and the constants defined.  Even if the compiler is smart enough to unscramble ad-hoc definitions, its harder to read through by a human.

stevenj: I do think types and constants should be grouped together,  rather than spread around throughout the trait definition.

book/src/rust/rust_style_guide.md
```diff
@@ -0,0 +1,697 @@
...
+impl Configuration {
+    const SPECIAL_CONSTANT: usize = 12345;
+    const ANOTHER_SPECIAL_CONSTANT: usize = 9876;
+}
+```
+
+#### Errors
+
+TODO: - @minikin
+
```
stevenj, line 578: Needs discussion, but we should Either:
1. always manually define custom errors.
2. Choose and then consistently use 1 library to simplify custom error definitions (such as https://crates.io/crates/snafu). 

This library is an example, not a recommendation.

We should not do a mix of the above with whatever library the developer feels like using at the time.

book/src/rust/rust_style_guide.md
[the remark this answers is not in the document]
stevenj: Possibly not for this document, but we should have a list of crates we use by preference, so that when someone needs a crate they look at the curated list and they use the one in that, that provides what they need.  Then, if they add a new crate, they should update that list with the crate, and what they use it for.  If they need to add a new crate, it should entail a brief discussion on slack, especially when there are options, before diving in.  

I find in rust there are usually 2 or more crates to do the same thing, of varying quality.  Its really easy to end up using different crates for the same things all over the place in a big code base.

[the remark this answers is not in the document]
stevenj: And in review,  this should be checked.  
1. Did a new crate get added as a dependency thats not in the list of "used crates", if so why.
2. Is someone re-inventing the wheel (ie, writing code for something when a perfectly usable crate already exists which does it.)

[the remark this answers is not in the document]
Mr-Leshiy: I think we can add a new section or a new check item into the `pull_request_template.md`, where it is important to add which new dependencies are coming with this new PR.

Mr-Leshiy:
I agree with the @cameron1024 comment that `rustfmt` and `clippy` should be the source of truth for us, so I think sections like `whitespaces`, `Spaces` etc. we can remove from this guideline.

book/src/rust/rust_style_guide.md
Mr-Leshiy: I would leave this ordering, but maybe mark it as a recommended in our style guide (potentially can do for any other ordering rules, mark them as a preferable not a mandatory)

book/src/rust/rust_style_guide.md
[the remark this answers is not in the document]
Mr-Leshiy: Created https://github.com/input-output-hk/catalyst-core/pull/207

book/src/rust/rust_style_guide.md
FelipeRosa: https://crates.io/crates/thiserror is pretty good (and popular).

minikin: @saibatizoku Do you remember we discussed an error guide for Rust a few weeks ago? I can't find that URL.

book/src/rust/rust_style_guide.md
```diff
@@ -250,3 +262,97 @@ When combined with the `include` macro, this can lead to pretty concise but also
...
+This allows the user to write:
+```rust
+match try_foo() {
+  Ok(foo) => println!("got a foo: {foo}"),
+  Err(FooError::Bar) => eprintln!("failed to bar"),
+  Err(FooError::Baz) => eprintln!("failed to baz"),
+}
+```
+
+#### Use `thiserror` for unrecoverable errors
```
FelipeRosa, line 350: Should this be `color_eyre` instead of `thiserror`?

FelipeRosa (approved):

cameron1024:
After discussing it in the technical discussion, we're in agreement that we can merge this

cameron1024 (approved):

minikin:
Thank you, everyone!

saibatizoku (approved):

PR #193, opened 2022-12-28
Update build and ci_tests workflows

minikin:
- Ignore changes in the `book` and `nix` folders in the `build` and `ci_tests` workflows.


jmgilman (approved):
Much needed change, thank you! 🙏

PR #194, opened 2022-12-28
Db sync mock setup

dkijania:
Introduced mock mode for db sync explorer which will create arbitrary in memory data based on simplified configuration send via rest. for more information please visit documentation for vitup, especially regarding snapshot definition


src/vit-testing/snapshot-trigger-service/src/client/mod.rs
```diff
@@ -1,190 +1,2 @@
 pub mod args;
 pub mod rest;
-
```
kukkok3, line 3: Has this file been deleted on porpuse?

kukkok3 (changes requested):
LGTM just a single comment to address

src/vit-testing/snapshot-trigger-service/src/client/mod.rs
dkijania: Yes , it was moved to integration tests since operations here working on top of snapshot service client. It caused issues when some library has dependency to snapshot service and voting tools and mainnet-tools. Code below is used integration tests only and break SPOR rule so i decided to move them to integration-tets

kukkok3 (approved):
LGTM

PR #198, opened 2023-01-03
use stable devshell for now

cameron1024:
a regression has snuck into nightly, this reverts it for now


jmgilman (commented):
.prettierrc shouldn't be getting checked in :)

Mr-Leshiy (approved):

PR #200, opened 2023-01-05
Make timestamps BIGINT and some other adjustments

FelipeRosa:
These were some issues I found while testing the DB schema against the current vit-ss build.

Converting timestamps to BIGINT is needed because vit-ss uses BIGINT to store timestamps and `EXTRACT(EPOCH FROM somecolumn)` gives a floating point number containing the Unix timestamp in its integral part.


minikin:
@FelipeRosa Can you explain briefly why do we need to have timestamps as BIGINT, please?

minikin (approved):

stevenj (approved):

PR #204, opened 2023-01-07
build: makes containers easier to troubleshoot

jmgilman:
Introduces several improvements to containers:

1. No longer sets the default container user to root (improve security)
2. Initializes the Nix database and adds the Nix CLI (with flake support). 
3. Adds the `nano` editor to allow modifying files on the container filesystem
4. Configures SSL so that additional packages can be fetched from `nixpkgs` at runtime

The result is that installing additional packages at runtime can now be achieved with:

```
nix profile install nixpkgs#vim
```

This greatly enhances the ability to debug images at runtime. The downside is that container images have grown by about ~200MB. I think this tradeoff is worth it for now, but once we have more stability, we can start using more optimized images.


stevenj (commented):
Is it possible to do this selectively,  so that production builds do not have it, but other development builds do?

jmgilman:
Yes, at the expense of increasing CI time since we'd have to produce two versions of every image. I do plan on eventually doing this, but in the meantime, we need more freedom on the images to increase our velocity with getting deployments working.

minikin (approved):

PR #205, opened 2023-01-07
build: adds migration support to servicing-station

jmgilman:
Allows the servicing station container to automatically run database migrations by setting the `MIGRATION` environment variable.

### DIFF
--- a/nix/lib/lib.nix
+++ b/nix/lib/lib.nix
@@ -69,7 +69,8 @@ in rec {
     pkgPath,
     cargoOptions ? [],
     nativeBuildInputs ? [],
-  }: let
+    ...
+  } @ args: let
     rootPkgCargo = readTOML (inputs.self + "/Cargo.toml");
     pkgCargo = readTOML (inputs.self + "/src/" + pkgPath + "/Cargo.toml");
     inherit (pkgCargo.package) name version;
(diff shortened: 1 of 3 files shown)


stevenj:
I do not want auto-migration.

Migration of DB schemas should always be intentional.  Not a side effect of a deployment.

All migrations should entail:
1. Bringing down EVERY service that is using the DB and schema.
2. When the DB is stable, FULL back up.
3. Migrate schema.
4. Run any tests that exist on the schema to ensure the migration was successful.
5. Redeploy all services that use the new schema.

Regardless of if there is 1 or more services,  the above should be the process.

Also,  vit-ss currently defines its schema.  But that will change immediately after the SVE.  For fund 10, vit-ss will use a common database that multiple services are using/updating and will not OWN its schema.  The DB/Schema will be their own thing.

stevenj (changes requested):
See comment regarding auto migrations of schemas in services.

jmgilman:
For clarity, this is only automatic insofar as the appropriate environment variable is set. By default, it will not be set, it's really just a flag that we can invoke to run a migration from inside of the VPC network. It's important to realize that:

1. The RDS instance is inaccessible from the internet
2. The only way to access it locally is to forward a connection using `socat` and `kubectl port-forward`
3. Applying migrations may be a semi-routine activity during testing periods when clusters are being set up and/or databases are being nuked. 

I don't mind keeping migrations relegated to a script at this point. The purpose of this change was to allow applying a migration without relying on local port-forwarding (i.e., just deploy the service with `MIGRATE` set). 

We can table this for now.

PR #209, opened 2023-01-10
move tracing and opentelemetry crates to workspace dependencies

saibatizoku:
adjust member manifests.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,5 +56,18 @@ members = [
 [workspace.dependencies]
 proptest = { git = "https://github.com/input-output-hk/proptest.git" }
 
+# Observability tools to instrument, generate, collect, and export telemetry data. 
+opentelemetry = { version = "0.18", features = ["rt-tokio"] }
+opentelemetry-otlp = "0.11.0"
+opentelemetry-semantic-conventions = "0.10.0"
+
+# Framework for instrumenting Rust programs to collect structured, event-based diagnostic information.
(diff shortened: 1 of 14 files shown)


cameron1024 (approved):

saibatizoku:
I've also set the full versions for the crates, but we should probably make use of a dependency bot to stay informed on updates.

cameron1024:
What's the motivation behind fully specifying the version numbers?

saibatizoku:
> What's the motivation behind fully specifying the version numbers?

No big motivation, just following what crates.io recommends in the "Installation" section. Plus, I figured that setting a new version will need to update Cargo.lock, and that should be a throttled move (maybe :crossed_fingers: when we set `workspace.version = "x.x.x"`, this will need to be updated at this point). But I have no strong feelings on this.

saibatizoku:
> > What's the motivation behind fully specifying the version numbers?
> 
> No big motivation, just following what crates.io recommends in the "Installation" section. Plus, I figured that setting a new version will need to update Cargo.lock, and that should be a throttled move (maybe crossed_fingers when we set `workspace.version = "x.x.x"`, this will need to be updated at this point). But I have no strong feelings on this.

I also think that there are projects out there that don't fully follow semantic versioning ... ehem ..., so it seems a bit less uncertain to specify the full version.

Mr-Leshiy (commented):
LGTM

PR #214, opened 2023-01-11
Remove transaction's expiration date filed validation

Mr-Leshiy:
# Description

Remove any transaction's expire date validations and checks. Also update chain-wallet-libs JS bindings, so during transaction building process it is unnecessary to provide this deprecated field.

## Type of change

- Refactoring
- This change requires a documentation update

## How Has This Been Tested?

As it is refactoring, just updated the current tests

## Checklist:


stevenj:
What impact does this have on the mempool?  
Are transactions stuck pending after this change?  
I didn't notice any mempool timeout handling or is that the subject of another PR?

src/chain-libs/chain-impl-mockchain/src/ledger/check.rs
```diff
@@ -146,13 +144,13 @@ pub enum TxVerifyError {
...
 }
 
-#[derive(Debug, Error, Clone, PartialEq, Eq)]
-pub enum TxValidityError {
-    #[error("Transaction validity expired")]
-    TransactionExpired,
-    #[error("Transaction validity is too far in the future")]
-    TransactionValidForTooLong,
-}
+// #[derive(Debug, Error, Clone, PartialEq, Eq)]
```
cong-or, line 147: is there a reason for commenting out `TxValidityError` ? Do we still need it?

src/chain-libs/chain-impl-mockchain/src/ledger/ledger.rs
```diff
@@ -931,14 +929,12 @@ impl Ledger {
                     return Err(Error::StakeDelegationSignatureFailed);
                 }
 
-                let (new_ledger_, _fee) =
-                    new_ledger.apply_transaction(&fragment_id, &tx, block_date)?;
```
cong-or, line 935: are there any potential adversarial implications of removing `BlockDate` . For example, it seems to be used to verify if the TX date is valid.

src/chain-libs/chain-impl-mockchain/src/ledger/tests/transaction_tests.rs
```diff
@@ -41,74 +40,7 @@ pub fn transaction_fail_when_255_outputs() {
             expected: 254,
             actual: 255
         }),
-        test_ledger.apply_transaction(fragment, BlockDate::first())
-    );
-}
-
-#[test]
-pub fn transaction_fail_when_validity_out_of_range() {
```
cong-or, line 49: should we have some type of mempool test to replace these? are TXs flushed or do they backlog?

Mr-Leshiy:
@stevenj At this PR I have removed a strict ledger validation for the transactions. So theoretically it is possible to submit into the blockchain "expired" transaction. Of course I have also modified an initial pool validation for the transactions, so at this moment it is also possible to submit into the pool expired transaction and this transaction will not be declined. 
**BUT** I have not touched any logic related to the cleaning the pool from the staled transactions, so pool still relies on this `expired_data` field in case of removing out dated transactions
https://github.com/input-output-hk/catalyst-core/blob/1cdee396d670a41dda614e1df3dc81fda4e5bfd9/src/jormungandr/jormungandr/src/fragment/pool.rs#L272 .
In the next PR I will modify this part as we have discussed before.

src/chain-libs/chain-impl-mockchain/src/ledger/ledger.rs
Mr-Leshiy: Expired date transaction's field is mostly used as an only way how to notify the user that a particular transaction will not submitted anymore into the blockchain. So this feature partially replaces transaction's error propagation which is missed.
Darius comment:
"
this is due the fact that we don't have error propagation through the network. Otherwise invalid fragments could congest the network. In our network, a follower node is accepting transaction and since it cannot mint a block just propagates it to the leader. If transaction is put in the block then follower node will receive update but as a block not transaction (once it receive new block , it unwraps all transactions from it and checks if any transaction currently residing in its mempool wasn't minted in the block. If yes then it updates mempool). However, there is a problem when leader rejects fragment. Leader node is silently rejecting fragments and not sending it back to follower, also failed transaction are not minted in the block. Therefore there was an idea of expiry time, after which follower node knows that transaction for sure will be rejected (even if it's correct but unfortunate enough to be pending all the time). It helps voting app to show user that transaction is failed, so it need to be resend, and since on each resend operation voting app sync with network it can fix a problem with wrong spending counter etc.
"

As it is discussed before and aligned that we want to remove this inconvenient for user field, and I assume later we are going to replace it with the proper error propagation mechanism.

src/chain-libs/chain-impl-mockchain/src/ledger/tests/transaction_tests.rs
Mr-Leshiy: I have not removed mempool tests related to the `expired_date` field they are still exist. As ledger validation has been removed these tests are unneeded anymore.

src/chain-libs/chain-impl-mockchain/src/ledger/ledger.rs
cong-or: ok, maybe we should create a ticket in relation to adding error propagation in this case.

cong-or (approved):
LGTM

PR #215, opened 2023-01-11
regen lockfile

cameron1024:
regenerates the lockfile, maybe this will make nix happy


minikin:
@cameron1024 Should we close this PR?

PR #216, opened 2023-01-11
Remove structopt

cameron1024:
This removes structopt in favour of clap v4.

The change is pretty much as simple as replacing `structopt` with `clap` wherever you see it, but there's a few more bits and pieces:
 - short command names now need a char literal instead of a string literal
 - subcommands need an explicit `#[clap(subcommand)]`
 - and a few extras


src/jormungandr/testing/mjolnir/src/mjolnir_lib/mod.rs
```diff
@@ -6,37 +6,38 @@ pub mod fragment;
...
-#[structopt(rename_all = "kebab-case")]
+#[derive(Parser)]
+#[clap(rename_all = "kebab-case")]
 pub enum MjolnirCommand {
     /// Passive Nodes bootstrap
     Passive(bootstrap::ClientLoadCommand),
     /// Explorer load
     Explorer(explorer::ExplorerLoadCommand),
     /// Fragment load
+    #[clap(subcommand)]
```
Mr-Leshiy, line 40: Maybe #[clap(subcommand)] is missing for other options

cong-or:
@cameron1024 I think [this script](https://github.com/input-output-hk/catalyst-core/blob/main/src/jortestkit/scripts/multiple-transactions-per-slot.sh#L39) may be the reason the test is failing.

> 'Command get: Short option names must be unique for each argument, but \'-h\' is in use by both \'host\' and \'help\' (call `cmd.disable_help_flag(true)` to remove the auto-generated `--help`)\

cameron1024:
Yep, didn't realise we used `-h` so heavily, but it's a shame to lose the help text. I think I'll just disable the help, but it's definitely unfortunate

cameron1024:
Addressing @Mr-Leshiy 's feedback here (since there's a few repeated comments)
 - `#[clap(subcommand)]` is used when the enum contains *another enum*, and it doesn't work for structs. Basically, you have no choice, you *must* add `subcommand` for enums, and you *must not* add `subcommand` for structs. That's why some enums have some variants with `subcommand` and other variants without
 - Clap v4 has a few more `Clone` bounds on data types than clap v2, so we had to add derive `Clone` to make it work. As for deriving `Copy`, I just did that on types where it "makes sense" (e.g. structs which just wrap a `[u8; N]`

Mr-Leshiy (dismissed):

Mr-Leshiy (approved):
LGTM

PR #219, opened 2023-01-12
Fix wallet-js publishing

Mr-Leshiy:
Update wallet-js package publishing, change wasm target from `nodejs` to `web`. Modify modules managing approach from `CommonJs` to `ES` standard.


minikin (approved):
LGTM.

PR #228, opened 2023-01-17
Update node metrics

Mr-Leshiy:
# Description

Replace `mempool_total_size` which shows current memory usage of the mempool with the `mempool_tx_count` which shows actual transactions amount inside the mempool

## Type of change
- Refactor


saibatizoku (dismissed):
Looks good to me.

saibatizoku (approved):

PR #229, opened 2023-01-17
small packages of adjustments for backend test services and new dockers for GVC client

dkijania:
This PR address need for dockers of catalyst-backend /  snapshot service / snapshot service wormhole. 

An example was provided in docker-compose file. Small fixes like accept configuration path from ENV or more verbose error reporting were done


dkijania:
Please hold on with review. I found one issue in the code i need to fix before handing it to review

stevenj (dismissed):
Looks Good.

PR #232, opened 2023-01-18
Update mempool functionality, get rid of the transaction's expiration date field processing

Mr-Leshiy:
# Description

Updated a mempool functionality, so now it works with the following algorithm:
1. Mempool has a specified `max_entries` field, which defines the maximum amount of transactions inside the pool
2. Transactions inside the pool are ordering as the deque `IndexedDeqeue` (first in first out).
3. You can add transactions into the pool using the `insert_all` function, which adds transaction in front the `queue` collection, if the pool if full (it contains a maximum amount of transactions) an oldest one is removed and a new one is added into the queue. When we are trying to submit more than the maximum amount of transactions, we are truncating an input collection, so we are actually submitting collection with the `max_entries` size.
4.  You pick up only the oldest transaction from the pool with the `remove_oldest` function, which takes transaction from the end of the deque.

## Type of change

- Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?

- [x] fragment::pool::internal::tests::overflowing_pool_should_remove_oldest_fragments
- [x] removed expired_fragment_should_be_rejected_by_passive_bft_node
- [x] removed expired_fragment_should_be_rejected_by_leader_praos_node  

## Checklist:


saibatizoku (approved):

src/jormungandr/jormungandr/src/fragment/pool.rs
```diff
@@ -188,8 +184,10 @@ impl Pool {
 
         let mut fragments = filtered_fragments.into_iter();
         let new_fragments = self.pool.insert_all(fragments.by_ref());
-        let count = new_fragments.len();
-        tracing::debug!("{} of the received fragments were added to the pool", count);
+        tracing::debug!(
+            "{} of the received fragments were added to the pool",
+            new_fragments.len()
+        );
```
FelipeRosa, line 190: You could do something like this to make this easier to find and index (eventually):

```suggestion
        tracing::debug!(
            count = %new_fragments.len(),
            "received fragments were added to the pool"
        );
```

src/jormungandr/jormungandr/src/fragment/pool.rs
```diff
@@ -553,95 +474,42 @@ pub(super) mod internal {
...
-            let max_fragments = self.max_entries - self.entries.len();
             fragments
                 .into_iter()
                 .filter(|(fragment, id)| {
                     if self.entries.contains(id) {
                         false
                     } else {
+                        if self.entries.len() >= self.max_entries && self.entries.len() != 0 {
+                            // Remove an oldest entry from the pool
+                            let (_, fragment) = self.entries.pop_back().expect("entry must exist");
```
FelipeRosa, line 485: The fragments are being inserted at the front so, to make it FIFO, this should use `pop_front` instead. Did I get this right?

FelipeRosa: Saw it at line 489.

FelipeRosa: Oh right, makes sense. My bad. Thanks for clarifying

saibatizoku (approved):

PR #239, opened 2023-01-25
docs: :bug: Rust code index rendered wrong; missing css.  

stevenj:
Fixed the rust code index which had its format broken when making it pass markdownlint rules.  It now has correct format, AND still passes markdownlint.
There was also CSS missing from the published docs which would lead to incorrect presentation.  This should fix that.
The last docs publication did not publish properly, so this is also a test to see if it works correctly, now.


utilities/mdbook-gen-summary/mdbook-gen-summary.py
```diff
@@ -0,0 +1,164 @@
+#!/usr/bin/env python3
+# Simple program to create an index for rust doc documentation, to be embedded inside an md-book
+# We do not use any non-standard packages, so this script should "just work" if
```
cameron1024, line 3: trailing "if", is there something that's meant to be here?

cameron1024 (approved):
Looks good. I can't really review the Python code, so feel free to get another opinion if you think it's appropriate, though given what it's doing it's probably not super correctness-sensitive

utilities/mdbook-gen-summary/mdbook-gen-summary.py
stevenj: fixed

cameron1024 (approved):

PR #241, opened 2023-01-25
Allow whitelisted peers to be specified using DNS names

FelipeRosa:
- Peers whitelist now uses `Multiaddr`
- Peer IP is used for reverse DNS lookup when checking if it's whitelisted
- Removed some blocking logic from inside the `FragmentProcessor` `Sink` implementation

PS1: `subscriptions.rs` is being hidden by GitHub because the diff is too big. Just a heads up
PS2: I was unable to test this locally inside the integration tests crate. So I left that out.


src/jormungandr/jormungandr/src/network/subscription.rs
```diff
@@ -288,6 +245,338 @@ impl Sink<net_data::Header> for BlockAnnouncementProcessor {
...
+                    }
+                    None => {
+                        tracing::debug!("no whitelist configured, bypassing check");
+
+                        true
+                    }
+                };
+
+                if should_send {
+                    let fragments = self.clear_buffered_fragments();
```
cong-or, line 504: why do we flush the buffer here?

src/jormungandr/jormungandr/src/network/subscription.rs
```diff
@@ -288,6 +245,338 @@ impl Sink<net_data::Header> for BlockAnnouncementProcessor {
...
+                    )
+                }
+            }
+        };
+
+        self.send_fragments_state = next_st;
+        ret
+    }
+
+    fn poll_flush_mbox(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Error>> {
```
cong-or, line 558: when does this happen? wondering if there any implications as the mbox is shared among nodes right?

src/jormungandr/jormungandr/src/network/p2p/comm.rs
```diff
@@ -465,23 +466,44 @@ pub struct PeerInfo {
     pub stats: PeerStats,
 }
 
+/// Future for executing async address lookup
+/// for the given node id from [Peers].
+pub struct GetNodeAddress {
```
cong-or, line 471: what happens when this can't resolve?

src/jormungandr/jormungandr/src/network/subscription.rs
FelipeRosa: I guess `clear_buffered_fragments` is kind of a bad name. This is what was already done in this part. It just empties the fragments buffer and sends all of them through the message box.

https://github.com/input-output-hk/catalyst-core/pull/241/files/d284ff271c9c130b5833e64a025b4528125566eb#diff-c9c45cadddc07944c423cf18d0c0775c7fda7d1c69e0ec690e21fa8bcda5e0b8L375-L378

src/jormungandr/jormungandr/src/network/p2p/comm.rs
FelipeRosa: That's right. I just kept the same behavior as before: https://github.com/input-output-hk/catalyst-core/pull/241/files/d284ff271c9c130b5833e64a025b4528125566eb#diff-c9c45cadddc07944c423cf18d0c0775c7fda7d1c69e0ec690e21fa8bcda5e0b8L316-L330

cong-or (approved):
LGTM

PR #243, opened 2023-01-26
Voting tools rewrite

cameron1024:
This PR is a significant rewrite of the `voting_tools_rs` crate. A summary of what's changed:
 - all of the validation logic
 - many of the type definitions (to more closely match what actually counts as a "valid registration") - many of these will be moved into a more fundamental "types" crate once 
 - there is new logic for generating fake registrations and signing them, which we use for property testing
 - Doc comments expanded
 - Cip 15's registration test vector is included, but many of the tests that try to validate that it's correct fail - however, this is because the cip15 vector itself is broken, so they **should** fail. I've left those tests in but ignored them so we can reuse them when we get some proper test vectors


stevenj (dismissed):
The code quality looks OK to me.
As a standalone tool I think we can just merge this and then get into heavy testing comparative against the old haskell tool.
I don't think this needs an extensive review.

Mr-Leshiy (dismissed):
LGTM

stevenj (approved):

PR #244, opened 2023-01-26
generic api sketch

cong-or:
Discuss strategies for generic API abstraction.


cong-or:
@stevenj 

to get started, what do you think of _store_ ? Basically removing blockchain_metadata param

 the rough sketch defined in mdbook
_mdbook_ serve

PR #245, opened 2023-01-26
Fix wallet-js interface, fix bug with the private casting

Mr-Leshiy:
Fixing wallet-js interfaces to be correspond to the CIP-62.
Also fix a bug related to the private vote cast generation


stevenj (approved):
Looks good.

PR #247, opened 2023-01-26
ci: Add Postgres container service for vit-ss tests

FelipeRosa:


.github/workflows/ci_tests.yml
```diff
@@ -174,6 +187,9 @@ jobs:
         run: cargo nextest archive --archive-file nextest-archive.tar.zst
 
       - name: Run Catalyst Core tests
+        env:
+          TEST_DB: postgres
+          TEST_DATABASE_URL: postgres://postgres:123456@localhost
```
FelipeRosa, line 192: This can be done better later. It seems I named the test db env vars differently in different parts of vit-ss

.github/workflows/ci-tests-n-coverage.yml
```diff
@@ -24,6 +24,19 @@ jobs:
     name: Unit Tests
     runs-on: ubuntu-latest
     needs: [integration-tests]
+    services:
+      postgres:
+        image: postgres:14
+        env:
+          POSTGRES_PASSWORD: 123456
```
FelipeRosa, line 31: Should this be hidden? I don't think leaving it like this is a problem but I can be missing something here.

saibatizoku: > Should this be hidden? I don't think leaving it like this is a problem but I can be missing something here.

Maybe just leave it blank?

stevenj (approved):
LGTM

.github/workflows/ci_tests.yml
stevenj: OK

PR #250, opened 2023-01-27
refactor: Remove SQLite support from vit-ss

FelipeRosa:
This PR removes SQLite support from vit-ss:
- Remove `q` and `execute_q` macros which were being using to run DB queries with different diesel types depending on the db connection type
- Removes checks for existence of SQLite db file
- Removes SQLite migration files
- Tests only run for Postgres now, so db url env vars are required to run (https://github.com/input-output-hk/catalyst-core/pull/247)
- I also changed vitup to stop using SQLite


.github/workflows/ci-tests-n-coverage.yml
```diff
@@ -65,7 +65,7 @@ jobs:
 
       - name: Run unit tests
         env:
-          TEST_DB: postgres
+          TEST_DATABASE_URL: postgres://postgres:postgres@localhost
```
FelipeRosa, line 68: I removed the duplicate test db env vars.

stevenj (approved):
Looks ok.

PR #251, opened 2023-01-28
Discontinue evm feature

saibatizoku:
This discontinues the `evm` feature available in crates within the workspace. It also removes `chain-evm` from the workspace. The last commit with the crate is stored as the git tag: https://github.com/input-output-hk/catalyst-core/tree/chain-evm-discontinued

This reduces the build time considerably.


cameron1024 (dismissed):
-7000 lines 🎉 LGTM

saibatizoku:
@jmgilman any idea why the nix CI test for jormungandr is failing?

minikin (changes requested):
@saibatizoku Should we create a tag or branch of the system with the state before evm things were removed 🧐?

saibatizoku:
> @saibatizoku Should we create a tag or branch of the system with the state before evm things were removed monocle_face?

Yeah, @minikin , I added a tag: https://github.com/input-output-hk/catalyst-core/tree/chain-evm-discontinued

minikin (approved):

jmgilman:
Sorry I missed the ping on this. We [build with prometheus](https://github.com/input-output-hk/catalyst-core/blob/main/nix/jormungandr/packages.nix#L15) by default because we rely on it for metric scraping. It appears something is broken when that feature is enabled.

PR #267, opened 2023-02-13
NPG 5113 adjust vit ss tests to use the new db schema when setting up the db

FelipeRosa:
- ci: Add collate to Postgres test instance
- refactor: Adjust election-db crate vit-ss compatibility views
- refactor: Adjust test db creation to use election-db crate schema
- refactor: Adjust vit-ss tests to comply with the revised db schema
- fix: Add election-db crate as vit-ss lib to fix Nix builds
- refactor: fix more vit-ss tests
- docs: Add election-db book folder index file


stevenj (approved):
LGTM

PR #273, opened 2023-02-17
Add db seed subcommand to ideascale-importer

FelipeRosa:
This PR adds a command to generate random DB seed data for the new vit-ss schema while still being compatible with the old schema.


stevenj (approved):
LGTM

PR #274, opened 2023-02-20
fix: Fix epoch_instant function, remove unreachable!() section 

Mr-Leshiy:
Fixed epoch_instant function, which was the reason of node failure, remove unreachable!() section and return just a current time


src/jormungandr/jormungandr/src/leadership/process.rs
```diff
@@ -150,14 +150,10 @@ impl Module {
...
-            Err(err) => {
+            Err(_) => {
                 // only possible if `epoch_time` is earlier than now. I.e. if the next
                 // epoch is in the past.
-
-                unreachable!(
-                    "next epoch is in the past. This is not possible, but it seems it append. {}",
-                    err
-                )
+                Ok(Instant::now())
```
stevenj, line 156: I think we should log a warning when we see this.
And print the epoch_time we expected, and the system_time we checked against.
This could also happen with clock skew.
On reflection, I think this comparison is just wrong TBH and SystemTime is not what we should be using.
`epoch_time` should not as far as I can see even be a type of SystemTime.  They are incomparable and we should not assume SystemTime means anything by itself.

To be correct, epoch_time should be converted to a known time base.
Real clock time should be converted to the same time base.
And then compared.
And they should not need to be exact, but within a reasonable window because clock skew is a real thing.

Quote from docs:
```
a SystemTime cannot be directly inspected
```
That means that what is inside SystemTime is implementation dependent.  epoch_time should be a known value or convertible to one.

stevenj (changes requested):
I think epoch_time as SystemTime is wrong and needs serious rethinking.

stevenj (dismissed):
I have looked harder and this is ok for this issue.
Looking at SystemTime issues can wait

Mr-Leshiy:
@stevenj cool, I will add some additional logging at this place. and will merge this pr

stevenj (approved):

PR #275, opened 2023-02-21
build: adds support for resetting jormungandr persistent storage

jmgilman:


const-iohk (approved):
Yes, if reset file is present, erase all and start fresh. LGTM.

PR #237, opened 2023-01-20
Remove spending counter from Vote Cast transactions

saibatizoku:
Removes the spending counter checks from the vote cast transactions spending.

PR #278, opened 2023-02-23
fix: Updating voting tests

kukkok3:
Fixing tests for https://github.com/input-output-hk/catalyst-core/pull/237
Removing confirm_transaction will avoid updating the spending counter to check voting transaction works anyway


Mr-Leshiy (approved):
LGTM

PR #284, opened 2023-03-02
feat: NPG-5482 Initialize "Catalyst Event Data" service

Mr-Leshiy:
# Description

Add an initial implementation of the Catalyst Event Data service. Added axum service initialization with clap and tracing

### DIFF
--- a/src/cat-data-service/src/service/mod.rs
+++ b/src/cat-data-service/src/service/mod.rs
@@ -1,0 +1,26 @@
+use axum::Router;
+use std::net::SocketAddr;
+
+mod v0;
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("Cannot run service, error: {0}")]
+    CannotRunService(String),
+}
+
+#[tracing::instrument]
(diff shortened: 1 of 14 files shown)


saibatizoku (approved):

src/cat-data-service/src/service/mod.rs
```diff
@@ -0,0 +1,26 @@
...
+pub enum Error {
+    #[error("Cannot run service, error: {0}")]
+    CannotRunService(String),
+}
+
+#[tracing::instrument]
+pub async fn run_service(addr: &SocketAddr) -> Result<(), Error> {
+    tracing::info!("Starting service...");
+    tracing::info!("Listening on {}", addr);
+
```
stevenj, line 16: You should pass in the settings, i think.
In case we need to add settings that an endpoint would want to consume.
Either that, or we should store the settings in a `once_cell` or the like.  In which case the bind address can be read from their and we don't need a parameter here for it.  <-- I prefer this approach rather than passing settings around all over the place.

Either way, we should print out all settings here when the service starts.

stevenj (commented):
It looks good.
My comments are quite pedantic and nothing major.

PR #285, opened 2023-03-03
Fix snapshot tweaks

stevenj:
Fixes formats and things to align with catalyst_toolbox and the haskell snapshot tool


src/voting-tools-rs/src/logic/mod.rs
```diff
@@ -123,7 +123,7 @@ fn convert_to_snapshot_entry(
         .get(&stake_key)
         .ok_or_else(|| eyre!("no voting power available for stake key: {}", stake_key))?;
 
-    let voting_power = voting_power.clone();
+    let voting_power = voting_power.to_u128().unwrap_or(0);
```
cameron1024, line 126: Do we want to panic here? Admittedly overflowing a u128 is unlikely, but it probably indicates a bug, and we should probably panic in that case

src/voting-tools-rs/src/validation/mod.rs
```diff
@@ -122,11 +122,14 @@ fn validate_signature(
...
+    let pub_key = Ed25519::public_from_binary(registration.stake_key.as_ref())
+        .map_err(|e| RegistrationError::StakePublicKeyError { err: e.to_string() })?;
+    let sig = Ed25519::signature_from_bytes(sig.as_ref())
+        .map_err(|e| RegistrationError::SignatureError { err: e.to_string() })?;
+
 
     match Ed25519::verify_bytes(&pub_key, &sig, &hash_bytes) {
         Verification::Success => Ok(()),
-        Verification::Failed => Err(RegistrationError::MismatchedSignature { hash_bytes }),
+        Verification::Failed => Ok(()), //Err(RegistrationError::MismatchedSignature { hash_bytes }),
```
cameron1024, line 133: There's a flag in `ValidationCtx` that we can use to turn off signature verification. Should we use that?

cameron1024 (approved):
Looks good - have we got any guidance about how big we should expect voting power to be? I've seen u64, u128, and BigDecimal, should we standardise on one?

src/voting-tools-rs/src/logic/mod.rs
stevenj: No, I don't want ANY avoidable panics in this tool.
Because the consumer is an automated process and it can not deal with it.
Things will just catastrophically break.
IF we calculated voting power down to Lovelace,  Maximum supply is 45,000,000,000 ADA * 1,000,000 Lovelace.
Which is way smaller than a U64.
I don't like bigdecimal because it feels like deciding not to understand the limits of the system your building, and all systems have limits.  its also a "real" number and we don't have fractional components in voting power.  We should limit ourselves to integers, and if someone wants voting power down to "Lovelace" then that the base unit.

So U64 would be totally safe.
But U128 is trivial and fast to deal with on any modern computer, even SBC's like the raspberry PI.
So, I would be totally comfortable with all processing of these numbers being u128.
No one is making a currency with U128 worth of voting power.
Checking ranges and validity is outside the scope of this tool.
If we use U128, then this conversion is unnecessary.
The only way this could fail to convert (I guess) is if its some really small value < 0.  Not sure what bigdecimal does on that conversion to u128.  But in any event, 0 should be the result.
I would not change it for this tool,  not would I change BigDecimal because it works.  But we should not use it anywhere else without very very good reason.

src/voting-tools-rs/src/validation/mod.rs
stevenj: I would rather not, without a concrete use case for it.
We should be emitting all registrations that fail signature validation to the errors.json file anyway, so nothing is lost.
I only did this so I can test.

src/voting-tools-rs/src/error/mod.rs
```diff
@@ -52,6 +52,22 @@ pub enum RegistrationError {
 
     #[error("stake key has wrong type: {0}, expected 14 or 15")]
     StakeKeyWrongType(u8),
+
+    #[error(
```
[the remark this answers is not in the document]
stevenj: Agree, but for this tool they are sufficient.

PR #287, opened 2023-03-07
fix: Minor migration modifications

saibatizoku:
Couple of fixes to get migrations to work, and change the primary key definition from the `voting_node` table.


saibatizoku:
Closing this, it duplicates #288

PR #289, opened 2023-03-07
feat: Voting node service deploy leaders

saibatizoku:
Implements logic to deploy jormungandr nodes after retrieving `block0` information from Event DB.


stevenj (approved):
LGTM

PR #294, opened 2023-03-16
feat: Voting power snapshot automation script

FelipeRosa:
In this PR:

- Started using Poetry for dep management
- Added import snapshot command but still missing some parts of the workflow:
   - Couldn't test it with the working snapshot_tool (field names will probably change)
   - Couldn't get dreps from GVC API

I'm opening this as it is to be able to work on other stuff while I figure out the missing parts.

PS: I'll rename the tool in a future PR (it's not only ideascale importer anymore).


utilities/ideascale-importer/ideascale_importer/snapshot_importer/importer.py
```diff
@@ -0,0 +1,287 @@
...
+    async def _run_snapshot_tool(self):
+        snapshot_tool_cmd = f"{self.config.snapshot_tool.path} --db-user {self.config.dbsync_database.user} --db-pass {self.config.dbsync_database.password} --db-host {self.config.dbsync_database.host} --db {self.config.dbsync_database.db} --min-slot 0 --max-slot {self.snapshot_tool_max_slot} --out-file {self.raw_snapshot_tool_file}"
+        await run_cmd(self.console, "snapshot_tool", snapshot_tool_cmd)
+
+    async def _run_catalyst_toolbox_snapshot(self):
+        # Could happen when the event in the database does not have these set
+        if self.catalyst_toolbox_min_stake_threshold is None or self.catalyst_toolbox_voting_power_cap is None:
+            raise RunCatalystToolboxSnapshotFailed("min_stake_threshold and voting_power_cap must be set")
+
+        catalyst_toolbox_cmd = f"{self.config.catalyst_toolbox.path} snapshot -s {self.raw_snapshot_tool_file} -m {self.catalyst_toolbox_min_stake_threshold} -v {self.catalyst_toolbox_voting_power_cap} --output-format json {self.catalyst_toolbox_out_file}"
```
stevenj, line 151: Line length.
Suggest breaking CLI command on options:
```
        catalyst_toolbox_cmd = (
        f"{self.config.catalyst_toolbox.path} snapshot"
        f" -s {self.raw_snapshot_tool_file}"
        f" -m {self.catalyst_toolbox_min_stake_threshold}"
        f" -v {self.catalyst_toolbox_voting_power_cap}"
        f" --output-format json {self.catalyst_toolbox_out_file}"
    )
```

stevenj (changes requested):
The comments for line length and sleeping in loops applies in multiple places.  I only commented in one place to show the idea.

If changing from marshmallow to pydantic is a big lift,  then we don't need to do it.
But, if its just changing the library and a couple of tweaks, we should.

stevenj (approved):
Looks good

PR #300, opened 2023-03-29
feat: NPG-5482 Connect EventDB to the service, add db queries

Mr-Leshiy:
# Description

Connect EventDb to the cat-data-service. Update EventDb implementation, add snapshot queries, add serialized snapshot types etc. 
This is a first part, final fixes and tests would be in the further PRs


src/event-db/src/types/snapshot.rs
```diff
@@ -0,0 +1,155 @@
...
+    {
+        struct SnapshotVersionVisitor;
+        impl<'de> serde::de::Visitor<'de> for SnapshotVersionVisitor {
+            type Value = SnapshotVersion;
+            fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
+                formatter.write_str(
+                    r#"Expect one of the following options: "latest", "25", "Fund 10" etc."#,
+                )
+            }
+            fn visit_borrowed_str<E>(self, v: &'de str) -> Result<Self::Value, E>
```
cameron1024, line 38: I think this can be `visit_str` instead of `visit_borrowed_str` - we only need the `borrowed` variant if we require that the lifetime of the str is long enough for it to live in the final struct (e.g. if we had a struct like `struct Foo<'a> { inner: &'a str }`.

But since we don't have that, we can use `visit_str` which has a lifetime which is only guaranteed to last for the body of the function, which is enough for our needs

src/event-db/src/types/snapshot.rs
```diff
@@ -0,0 +1,155 @@
...
+    pub voting_power_saturation: f64,
+}
+
+#[derive(Serialize, Clone)]
+pub struct Voter {
+    pub voter_info: VoterInfo,
+    #[serde(serialize_with = "serialize_systemtime_as_rfc3339")]
+    pub as_at: SystemTime,
+    #[serde(serialize_with = "serialize_systemtime_as_rfc3339")]
+    pub last_updated: SystemTime,
```
cameron1024, line 71: I have a slight preference for `DateTime<Utc>` over `SystemTime`, because of the API it provides on the Rust side, but it's not a huge deal

src/event-db/src/queries/snapshot.rs
```diff
@@ -0,0 +1,122 @@
...
+
+        Ok(Voter {
+            voter_info: VoterInfo {
+                delegations_power: voter.try_get("delegations_power")?,
+                delegations_count: voter.try_get("delegations_count")?,
+                voting_power_saturation: if total_voting_power_per_group as f64 != 0_f64 {
+                    voting_power as f64 / total_voting_power_per_group as f64
+                } else {
+                    0_f64
+                },
```
cameron1024, line 67: style nit: can we pull this out into its own calculation just above line 59

src/event-db/src/types/snapshot.rs
```diff
@@ -0,0 +1,155 @@
...
+        let voter = Voter {
+            voter_info: VoterInfo {
+                voting_power: 100,
+                voting_group: "rep".to_string(),
+                delegations_power: 100,
+                delegations_count: 1,
+                voting_power_saturation: 0.4,
+            },
+            as_at: SystemTime::UNIX_EPOCH,
+            last_updated: SystemTime::UNIX_EPOCH,
```
cameron1024, line 124: Does this mean "not updated"? If so, maybe an `Option<SystemTime>` (or `Option<DateTime<Utc>>`) would be clearer?

src/event-db/src/lib.rs
```diff
@@ -22,6 +20,19 @@ const DATABASE_URL_ENVVAR: &str = "EVENT_DB_URL";
...
 
+/// Event database errors
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error(" Schema in database does not match schema supported by the Crate. The current schema version: {was}, the schema version we expected: {expected}")]
+    MismatchedSchema { was: i32, expected: i32 },
+    #[error(transparent)]
+    TokioPostgresError(#[from] tokio_postgres::Error),
+    #[error(transparent)]
+    TokioPostgresRunError(#[from] RunError<tokio_postgres::Error>),
```
cameron1024, line 31: This error type seems like it should be database agnostic (i.e. it appears in the trait, which is the same regardless of whether we use postgres, mongo, some in-memory thing, etc).

If that's the case, it might be worth having a single `Unknown(Report)` variant that is for "random database errors that we can't handle" (like connection errors), that way we're not leaking implementation details to the caller

cameron1024 (commented):
Looking good, just a few suggestions

src/event-db/src/types/snapshot.rs
```diff
@@ -0,0 +1,155 @@
...
+}
+
+#[derive(Serialize, Clone)]
+pub struct Voter {
+    pub voter_info: VoterInfo,
+    #[serde(serialize_with = "serialize_systemtime_as_rfc3339")]
+    pub as_at: SystemTime,
+    #[serde(serialize_with = "serialize_systemtime_as_rfc3339")]
+    pub last_updated: SystemTime,
+    pub r#final: bool,
```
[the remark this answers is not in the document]
Mr-Leshiy: I have followed just a API specification of these types https://github.com/input-output-hk/catalyst-core/blob/d2aabde031464526ab2fbaf09f1ef4a60ca81776/book/src/07_web_api/openapi/vitss_v0.yaml#L1247. 
But of course we can rename it, but it this case we will need to add serde(rename =).

src/cat-data-service/src/cli.rs
```diff
@@ -1,15 +1,16 @@
...
 use tracing::subscriber::SetGlobalDefaultError;
 
 #[derive(thiserror::Error, Debug)]
 pub enum Error {
     #[error(transparent)]
-    ServiceError(#[from] service::Error),
+    Service(#[from] service::Error),
     #[error(transparent)]
-    LoggerError(#[from] SetGlobalDefaultError),
+    Logger(#[from] SetGlobalDefaultError),
```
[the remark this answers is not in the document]
Mr-Leshiy: But this error will be handled here https://github.com/input-output-hk/catalyst-core/blob/d2aabde031464526ab2fbaf09f1ef4a60ca81776/src/cat-data-service/src/main.rs#L11, and if something will goes wrong it will definitely panics with the appropriate error. Or you think I should explicitly panic right here ?

src/event-db/src/types/snapshot.rs
Mr-Leshiy: No no no, it is not possible to make it optional, because it is a correspondance field to this one from the db schema https://github.com/input-output-hk/catalyst-core/blob/d2aabde031464526ab2fbaf09f1ef4a60ca81776/src/event-db/migrations/V6__snapshot_tables.sql#L9. Here in tests I just put some test values to test serde

src/event-db/src/queries/snapshot.rs
Mr-Leshiy: fixed

src/event-db/src/types/snapshot.rs
Mr-Leshiy: fixed

src/event-db/src/types/snapshot.rs
```diff
@@ -0,0 +1,155 @@
...
+
+    #[test]
+    fn snapshot_version_json_test() {
+        let snapshot_versions = vec![
+            SnapshotVersion::Latest,
+            SnapshotVersion::Number(10),
+            SnapshotVersion::Name("Fund 10".to_string()),
+        ];
+        let json = serde_json::to_string(&snapshot_versions).unwrap();
+        assert_eq!(json, r#"["latest","10","Fund 10"]"#);
```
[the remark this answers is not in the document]
Mr-Leshiy: fixed

src/cat-data-service/src/state.rs
```diff
@@ -0,0 +1,14 @@
+use crate::cli::Error;
+use event_db::{queries::snapshot::SnapshotQueries, schema_check::SchemaVersion, EventDB};
+
+pub struct State<EventDB: SnapshotQueries> {
```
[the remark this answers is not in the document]
Mr-Leshiy: fixed

src/cat-data-service/src/state.rs
```diff
@@ -0,0 +1,14 @@
...
+    pub event_db: EventDB,
+}
+
+impl State<EventDB> {
```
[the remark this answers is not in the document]
Mr-Leshiy: fixed

src/event-db/src/types/snapshot.rs
Mr-Leshiy: Fixed

src/event-db/src/lib.rs
Mr-Leshiy: Fixed

src/cat-data-service/src/cli.rs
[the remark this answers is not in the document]
cameron1024: I think we can just use `logger::init().unwrap()` - if it returns an error, we're going to panic anyways, just panicking here means can keep the error enum uncluttered for the error variants we care about

src/event-db/src/types/snapshot.rs
cameron1024: Ahh makes sense 👍

src/event-db/src/types/snapshot.rs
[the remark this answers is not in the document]
Mr-Leshiy: fixed

src/cat-data-service/src/cli.rs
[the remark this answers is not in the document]
Mr-Leshiy: fixed

src/cat-data-service/src/service/v0/fund.rs
```diff
@@ -1,43 +1,30 @@
...
+        .route("/fund", get(fund_exec))
+        .route("/fund/:id", get(fund_by_id_exec))
 }
 
-async fn fund_exec<State: FundDb>(state: Arc<State>) -> String {
+async fn fund_exec() -> String {
     tracing::debug!("fund_exec");
 
-    let current_fund = state.get_current_fund();
+    let current_fund: Fund = Default::default();
```
cameron1024, line 14: nit: could be `Fund::default()`

src/cat-data-service/src/service/v0/snapshot.rs
```diff
@@ -1,52 +1,62 @@
...
+        stake_public_key
+    );
 
-    let delegator = state.get_delegator(event, stake_public_key);
-    serde_json::to_string(&delegator).unwrap()
+    let delegator = state
+        .event_db
+        .get_delegator(event, stake_public_key)
+        .await?;
+    Ok(serde_json::to_string(&delegator).unwrap())
```
cameron1024, line 61: We don't need to do the string conversion here - we can just return `Json(delegator)` (`Json` comes from axum)

cameron1024 (approved):
A couple of small things I didn't see first time round, but looks good :grin:

src/cat-data-service/src/service/v0/fund.rs
Mr-Leshiy: fixed

Mr-Leshiy: Fixed

PR #303, opened 2023-03-31
feat: Add 'ideascale-importer' script to ideascale_importer package

saibatizoku:
Adds an executable script `ideascale-importer` to the `ideascale_importer` python package in `utilities/ideascale-importer`.

### DIFF
--- a/utilities/ideascale-importer/pyproject.toml
+++ b/utilities/ideascale-importer/pyproject.toml
@@ -44,6 +44,9 @@ flake8 = "6.0.0"
 mypy = "0.991"
 mypy-extensions = "0.4.3"
 
+[tool.poetry.scripts]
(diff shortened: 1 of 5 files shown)


utilities/ideascale-importer/pyproject.toml
```diff
@@ -44,6 +44,9 @@ flake8 = "6.0.0"
 mypy = "0.991"
 mypy-extensions = "0.4.3"
 
+[tool.poetry.scripts]
+ideascale-importer = "ideascale_importer.app:app"
```
saibatizoku, line 48: I kept the `app` name, but it feels like something like `ideascale_importer.app:cli` could better describe what's going on.

FelipeRosa (approved):

PR #304, opened 2023-04-03
Feature/npg 5770 rationalized catalyst 1.0 backend api

stevenj:


kickloop:
@stevenj I am still going through the details and I will leave some comments. So far all looks good.

I just had a thought. We use Blockfrost API for GVC to verify the transactions, specifically Voter registration and delegations. If we could use some of the new Catalyst endpoints, perhaps develop a new one, if needed, and get rid of the Blockfrost completely, that would be perfect. What do you think?

stevenj:
> @stevenj I am still going through the details and I will leave some comments. So far all looks good.
> 
> I just had a thought. We use Blockfrost API for GVC to verify the transactions, specifically Voter registration and delegations. If we could use some of the new Catalyst endpoints, perhaps develop a new one, if needed, and get rid of the Blockfrost completely, that would be perfect. What do you think?
> 
> cc @minikin

Yes its better if we don't need to use blockfrost.  I also wonder in blockfrost is properly validating the txn against CIP-36 (such as checking the signature is correct).  You should be able to get this data from this API.

I will also be adding endpoints to give the voting power for any unregistered stake address, during the snapshot.  So if someone isn't registered you can still show them what their voting power will be.

And an endpoint to get errors, so you can show them if there are any errors later than their last valid registration (or historical registration data).

But we will implement these first.

src/event-db/migrations/V2__event_tables.sql
```diff
@@ -34,7 +34,8 @@ CREATE TABLE event
     committee_size INTEGER NOT NULL,
     committee_threshold INTEGER NOT NULL,
 
-    extra JSONB
+    extra JSONB,
+    cast_to JSONB
```
Mr-Leshiy, line 38: Didn't really get the idea of this field. Can you please a little bit elaborate :)

Mr-Leshiy (approved):

src/event-db/migrations/V2__event_tables.sql
stevenj: `cast_to` is where an event casts its votes.  The format of the embedded json is just an idea at the moment.  But I wanted to add this field while I was in there so when we want to run multiple jorm chains, or push votes to mainnet or our p2p network, we have somewhere we can configure it.

I made it an array so we can declare multiple destinations, so we can cast votes into all of them at the same time. (Ie, jorm and the p2p network).  Just by tweaking the config.

Processing of casting votes would just iterate this array and send the vote where it needs to go (or do it in parallel).

Mr-Leshiy: Got it ! thanks !
But what do you think to use here an enumeration type ?
https://www.postgresql.org/docs/current/datatype-enum.html

stevenj: Those kinds of types require schema changes which I am trying to minimize.
Schema changes are a lot harder to manage than data changes within the same schema.

stevenj: For example,  we will define jormungandr as a place to cast votes first.
And we can probably guess now what we would need.
Just a "Chain ID" and the fact this is Jormungandr.
But what do we need for voting on the athena network?  No idea,  I can guess but it will probably be wrong.
But with a JsonB type it doesn't matter,  I can add an athena type later,  tools that understand it can process it, others will ignore it (or they should).  In fact testing off jsonb types needs to include unexpected data to make sure the consuming services don;t do anything crazy.  The basic schema of these json types also needs to be defined well enough that they are  unambiguous.

stevenj: As we go, I will actually create proper json schema to define what should be in these JSONB fields.

Mr-Leshiy: Got it, make sense, thank you !

book/assets/theme/FontAwesome/fonts/fontawesome-webfont.svg
Mr-Leshiy, line 1: Really awesome font :)

Mr-Leshiy (approved):

PR #307, opened 2023-04-04
fix: Fix snapshot importer logic

FelipeRosa:


utilities/ideascale-importer/ideascale_importer/snapshot_importer/importer.py
```diff
@@ -103,7 +105,7 @@ async def _fetch_parameters(self):
             )
 
             row = await conn.fetchrow(
-                "SELECT slot_no FROM block WHERE time <= $1 ORDER BY time DESC LIMIT 1",
+                "SELECT time, slot_no FROM block WHERE time <= $1 ORDER BY time DESC LIMIT 1",
```
stevenj, line 108: On reflection, for the latest slot_no and time, I think the query should just be:
```sql
SELECT time, slot_no FROM block SORT BY slot_no DESC LIMIT 1;
```
Which will give you the latest slot number and time in dbsync.

This query is still valid for getting the slot number the snapshot needs to be run as at.

So you need two slot numbers:
1. The "as at" slot number which is when we align the registrations and voting power.
2. The "current" slot number which is the latest time and slot number recorded by dbsync.

Final is only related to the second.

Also, before running the snapshot, we should check IF the latest snapshot for the event in the DB is already marked final and not re-run it if it is.

stevenj (changes requested):

FelipeRosa:
@stevenj I made the logic for setting the `final` column use the latest block time and separated the queries for the snapshot `slot_no` and the latest block time. Is this what you meant?

utilities/ideascale-importer/ideascale_importer/snapshot_importer/importer.py
```diff
@@ -103,10 +107,9 @@ async def _fetch_parameters(self):
             )
 
             row = await conn.fetchrow(
-                "SELECT slot_no FROM block WHERE time <= $1 ORDER BY time DESC LIMIT 1",
+                "SELECT MAX(slot_no) as slot_no FROM block WHERE time <= $1 AND slot_no IS NOT NULL",
```
stevenj, line 110: This will get the slot of the deadline,  but ONLY when the deadline has passed.
Which means the next line will fail by raising the exception.
Instead,  IF the deadline has not passed, and this returns None, it should just default to the latest slot number.
So we end up with 3 time bands.

1. Before the deadline.  (snapshot_tool_max_slot == latest_slot). Not Final.
2. After the deadline and latest_slot_time < snapshot_start (snapshot_tool_max_slot = slot just before deadline). Not final.
3. After the deadline and latest_slot_time >= snapshot_start (snapshot_tool_max_slot = slot just before deadline). Final.

utilities/ideascale-importer/ideascale_importer/snapshot_importer/importer.py
```diff
@@ -115,11 +118,23 @@ async def _fetch_parameters(self):
...
             self.snapshot_tool_max_slot = row["slot_no"]
             logger.info(
-                "Got max_slot for max_time",
+                "Got block time parameters for snapshot_tool",
                 snapshot_tool_max_slot=self.snapshot_tool_max_slot,
                 max_time=self.config.snapshot_tool.max_time.isoformat()
             )
 
+            row = await conn.fetchrow(
+                "SELECT max(time) AS latest_time FROM block",
```
stevenj, line 127: So we should get the latest slot_no here, just in case the query above failed, and so we can use it as the slot_no we run the snapshot against.  The as_at in the db would then need to be the time from this query if it defaults here.

utilities/ideascale-importer/ideascale_importer/snapshot_importer/importer.py
```diff
@@ -233,11 +266,27 @@ async def _write_db_data(self):
...
+            logger.info(
+                "Setting snapshot final flag",
+                final=is_snapshot_final,
+                snapshot_start=self.snapshot_start_time.isoformat(),
+                latest_block_time=self.lastest_block_time.isoformat(),
+            )
+
         snapshot = models.Snapshot(
             event=self.event_id,
             as_at=datetime.utcnow(),
```
stevenj, line 287: as_at needs to be the actual times from the DB.
In the time before the deadline, `as_at` = the slot time we ran the snapshot against.
After the deadline `as_at`  is the time of the deadline.

utilities/ideascale-importer/ideascale_importer/snapshot_importer/importer.py
```diff
@@ -233,11 +266,27 @@ async def _write_db_data(self):
...
             last_updated=datetime.utcnow(),
```
stevenj, line 288: last_updated needs to be the time read into `lastest_block_time` which tells us what time dbsync was at when we ran the command.  The time on the local machine isn't important due to dbsync and local time not being synchronized.

utilities/ideascale-importer/ideascale_importer/snapshot_importer/importer.py
```diff
@@ -233,11 +266,27 @@ async def _write_db_data(self):
...
-            final=False,
+            final=is_snapshot_final,
             dbsync_snapshot_cmd=os.path.basename(self.config.snapshot_tool.path),
             dbsync_snapshot_data=snapshot_tool_data_raw_json,
```
stevenj, line 291: Because this is huge, and I also want to store the other artifacts, I changed the DB spec to brotli encode this data.
But make that a separate ticket to implement.

stevenj (changes requested):

stevenj:
> @stevenj I made the logic for setting the `final` column use the latest block time and separated the queries for the snapshot `slot_no` and the latest block time. Is this what you meant?

Close, but not quite.  See notes above.

On reflection I think we should add `as_at_slotno` and `last_updated_slotno` to the `snapshot` table.  The reason is there is no easy way to know what slot relates to the time and vice versa, and both pieces of information are important.

So, add those two fields to the schema for the `snapshot` table and push the schema change with this PR.
Also add updating them to the query above :)
Doesn't need a separate migration, just update the current field in the schema.

utilities/ideascale-importer/ideascale_importer/snapshot_importer/importer.py
FelipeRosa: `max_time` is used to get the `max_slot` parameter the importer passes to snapshot_tool. This query is getting the slot_no at (at most) that time. This is independent of the deadline.

The snapshot deadline is fetched from the event in the DB and stored in the `Importer.snapshot_start_time` [here](https://github.com/input-output-hk/catalyst-core/pull/307/files/41c99240f7a13ea80a30ded7355618b6672bc265#diff-8c922024c086e6e0c07f0c8d115ec575671ba9bb385b86b6e69ab41bbd08aa2cR146). This is then used [here](https://github.com/input-output-hk/catalyst-core/pull/307/files/41c99240f7a13ea80a30ded7355618b6672bc265#diff-8c922024c086e6e0c07f0c8d115ec575671ba9bb385b86b6e69ab41bbd08aa2cR269) to decide whether the snapshot is final.

stevenj: As discussed in slack,  this is an error.  There are 2 times in the DB.
`registration_snapshot_time` this is the deadline, and the slot just PRIOR to this time is the `max_slot` passed to the snapshot tool.  But we need to make sure this time has past (as far as dbsync is concerned), because if it hasn't we want to use the latest slot we have.

`snapshot_start` is named this because its when we used to run snapshot when there was only 1.  But now it means snapshot is final if we ran one and the latest slot time was after this time.

stevenj (changes requested):

utilities/ideascale-importer/ideascale_importer/snapshot_importer/importer.py
```diff
@@ -63,24 +64,35 @@ class WriteDbDataFailed(Exception):
...
-        self.catalyst_toolbox_min_stake_threshold: Optional[int] = None
-        self.catalyst_toolbox_voting_power_cap: Optional[float] = None
+        self.lastest_block_time: Optional[datetime] = None
+        self.latest_block_slot_no: Optional[int] = None
+        self.registration_snapshot_slot: Optional[int] = None
+        self.registration_snapshot_block_time: Optional[datetime] = None
+        self.registration_snapshot_time: Optional[datetime] = None
+        self.snapshot_start_time: Optional[datetime] = None
+        self.min_stake_threshold: Optional[int] = None
+        self.voting_power_cap: Optional[float] = None
```
FelipeRosa, line 93: Renamed these to match the names in event-db

utilities/ideascale-importer/ideascale_importer/snapshot_importer/importer.py
```diff
@@ -94,51 +106,97 @@ def __init__(
...
                 f"postgres://{self.config.dbsync_database.user}:"
                 + f"{self.config.dbsync_database.password}@{self.config.dbsync_database.host}"
                 f"/{self.config.dbsync_database.db}"
             )
 
+            # Fetch slot number and time from the block right before or equal the registration snapshot time
             row = await conn.fetchrow(
-                "SELECT slot_no FROM block WHERE time <= $1 ORDER BY time DESC LIMIT 1", self.config.snapshot_tool.max_time
+                "SELECT slot_no, time FROM block WHERE time <= $1 AND slot_no IS NOT NULL ORDER BY slot_no DESC LIMIT 1",
+                self.registration_snapshot_time,
```
FelipeRosa, line 165: This should always return the latest block right before registration shapshot time

utilities/ideascale-importer/ideascale_importer/snapshot_importer/importer.py
```diff
@@ -94,51 +106,97 @@ def __init__(
...
+            raise FinalSnapshotAlreadyPresent()
+
+        await conn.close()
+
     async def _fetch_parameters(self):
+        # Fetch event parameters
+        conn = await ideascale_importer.db.connect(self.database_url)
+
+        row = await conn.fetchrow(
+            "SELECT registration_snapshot_time, snapshot_start, voting_power_threshold, max_voting_power_pct FROM event WHERE row_id = $1",
```
FelipeRosa, line 124: Fetching registration_snapshot_time and snapshot_start

utilities/ideascale-importer/snapshot-importer-example-config.json
```diff
@@ -6,8 +6,7 @@
         "db": "database name"
     },
     "snapshot_tool": {
-        "path": "snapshot_tool",
-        "max_time": "2023-03-15T00:00"
```
FelipeRosa, line 10: This comes from `registration_snapshot_time`

stevenj (approved):

stevenj:
This all looks good now. 
Just needs the conflicts fixed.

stevenj (approved):

PR #308, opened 2023-04-05
feat: Create committee keysets

saibatizoku:
Finish implementing the correct committee creation logic.

Adds a lot of documentation, and improved code legibility.


services/voting-node/pyproject.toml
```diff
@@ -12,7 +12,7 @@ requires = ["poetry-core"]
 build-backend = "poetry.core.masonry.api"
 
 [tool.poetry.dependencies]
-python = "^3.10"
+python = "^3.11"
 # formatting
 black = "^23.1.0"
```
stevenj, line 17: `black`, `ruff` should be "development" dependencies because we don't need to deploy them.
See: https://python-poetry.org/docs/managing-dependencies#dependency-groups

src/event-db/migrations/V8__catalyst_automation.sql
```diff
@@ -28,18 +28,49 @@ COMMENT ON COLUMN voting_node.netkey IS
...
+    FOREIGN KEY(committee) REFERENCES tally_committee(row_id)
+);
+
+COMMENT ON TABLE committee_member IS 'Table for storing data about the tally committee members.';
+COMMENT ON COLUMN committee_member.row_id IS 'Unique ID for this committee member for this event.';
+COMMENT ON COLUMN committee_member.committee IS 'The committe this member belongs to.';
+COMMENT ON COLUMN committee_member.pubkey  IS 'Public key for the member wallet.';
+COMMENT ON COLUMN committee_member.seckey  IS 'Secret key for the member wallet.';
+COMMENT ON COLUMN committee_member.member_pk  IS 'Committee member public key';
+COMMENT ON COLUMN committee_member.member_sk  IS 'Committee member secret key';
```
stevenj, line 76: Are the secret keys being stored encrypted in the DB?
We need to be able to safely publish the DB data, so any "Secret" keys we store need to be encrypted.
We need a `SECRET_SECRET` env var to control the "Secret" encryption key.
We should use this method:
https://cryptography.io/en/latest/fernet/#using-passwords-with-fernet

You need to derive the key on every encryption and decryption, which needs the salt.
Salt needs to be random on every encryption.
Save the salt with the encrypted data in the DB.
For example,  the salt is 16 bytes binary.
Get the encrypted binary data, pre-pend it with the 16 byte salt, base 64 the result and store that base 64 string as the encrypted secret.

You should just make a function which does this, and constrain the encryption/decryption to the DB layer on read and write.

stevenj (changes requested):
Changes requested:
1. Encryption of secret values in the DB.
2. Schema comments
3. Development Dependency group

Docs should not be updated as per my comment in this PR, but make a ticket TODO it in a subsequent PR.

services/voting-node/development.md
```diff
@@ -8,14 +8,13 @@ This service acts as a wrapper around a `jormungandr`, interacting with the Even
 ## Requirements
```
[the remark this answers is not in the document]
saibatizoku: Thanks, I'll add a ticket for the documentation work that you describe.

services/voting-node/pyproject.toml
saibatizoku: Done.

src/event-db/migrations/V8__catalyst_automation.sql
saibatizoku: Got it, I've added the functions that do encrypt/decrypt for storage.

stevenj (approved):
LGTM

PR #310, opened 2023-04-07
feat(event-db): NPG-5753 GraphQL schema server and security definitions

stevenj:
Add GraphQL server for admin and security.


schema.graphql
Mr-Leshiy, line 1: Is it possible to split it into the separate files ?

src/event-db/setup/graphql-setup.sql
```diff
@@ -0,0 +1,173 @@
...
+    about           text,
+    email           text not null unique check (email ~* '^.+@.+\..+$'),
+    password_hash   text not null,
+    created_at      timestamp default now()
+);
+
+-- Create default ADMIN user.
+
+insert into private.admin_account (first_name, last_name, role, about, email, password_hash) values
+  ('Admin', 'Default', 'cat_admin', 'Default Admin User', 'admin.default@nowhere.io', crypt('CHANGE_ME', gen_salt('bf')));
```
cong-or, line 56: should there be a larger salt?

stevenj: gen_salt generates a salt with the 'bf' type.
See: https://www.postgresql.org/docs/current/pgcrypto.html#PGCRYPTO-CRYPT-ALGORITHMS
its a 128-bit salt.

schema.graphql
stevenj: Its autogenerated.

cong-or (approved):
LGTM

PR #314, opened 2023-04-13
feat: NPG-5816 /api/v1/event/{id}/objectives endpoint (part 1)

Mr-Leshiy:
# Description

It is a first part of the `/api/v1/event/{id}/objectives` endpoint implementation. This Pr contains only event-db queries and fixes, implementation of the new objective's types. 

@stevenj pls take a close look at the db schemas updates.


stevenj (approved):
I checked the schema changes, and they look OK.
Everything else looks OK.

PR #316, opened 2023-04-14
Kill openssl

cameron1024:
Removes:
 - openssl
 - blockfrost stuff
 - db-sync-explorer


stevenj:
This is good.
I made 1extra removal.  `vitup` was importing `ipfs-api` which itself had a number of dependencies.  It was not however used by `vitup`.  The dependency tree now looks a lot more reasonable.

PR #318, opened 2023-04-18
feat: Automate ideascale importer in voting node

saibatizoku:
Add ideascale-importer utility as a dependency for voting-node.
Implement committee generation logic.
Improvement to documentation.
Secret key encryption/decryption.
Committee key generation per event.


stevenj (approved):
LGTM

PR #320, opened 2023-04-19
feat: Feature/historic data

stevenj:
Adds F0 to F4 Historical Data
These Funds are ready to review and merge.
This found a few issues with the DB schema, which I have also addressed.  
Check them because they may require changes to the cat-data-service to accommodate them.


Makefiles/db.toml
```diff
@@ -17,7 +17,7 @@ category = "db"
 script_runner = "@shell"
 script = '''
 cd ${CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY}/src/event-db
-psql -e postgres -f setup/dev-db.sql ${@}
+psql -e -U postgres -f setup/dev-db.sql ${@}
```
Mr-Leshiy, line 20: better not to add this option here. As we are using ${@} argument we can pass it like this
```
cargo make local-event-db-init -U postgres
```

Mr-Leshiy (approved):

saibatizoku (approved):
LGTM.

Makefiles/db.toml
stevenj: We are looking at improving CI/CD let's fix this issue when we do that.
The `-U` fixes the original file because it's broken without it.

PR #322, opened 2023-04-20
feat: NPG-5816 /api/v1/event/{id}/objectives endpoint (part 2)

Mr-Leshiy:
Last part of the /api/v1/event/{id}/objectives endpoint implementation.
At this moment `ballot: GroupBallotType` field is omitted and mocked data just returned. It will be fixed later after deep investigation on the `voteplan` or `objective` tables.


saibatizoku (commented):
LGTM.

PR #327, opened 2023-04-25
fix(event-db): Parameterize Event DB Init SQL scripts to make easier to use in production

stevenj:
This adds parameters to the DB init scripts so they can be configured at run time with different DB/User and password settings.
Make the built in dev options use the new parameters.


stevenj:
> Mostly there.
> 
> ```
> psql:/tmp/tmp.Fz50QwNx8y/setup.sql:85: ERROR:  role "cat_admin" cannot be dropped because some objects depend on it
> psql:/tmp/tmp.Fz50QwNx8y/setup.sql:98: ERROR:  role "cat_anon" cannot be dropped because some objects depend on it
> ```
> 
> Maybe just move the drops to the top instead of grouping them together? Then you can handle the correct order. This should make it idempotent.

Done

jmgilman (approved):
LGTM. Thanks!

PR #328, opened 2023-04-25
Feature/cat data service impl

kickloop:
I've added few more details to event-db/Readme


minikin (changes requested):
Looks like you accidentally added some unnecessary files.

kickloop:
> Looks like you accidentally added some unnecessary files.

Ah... its a linter or smth... I was running late on a meeting so I just pushed my changes. The only file I've changed is the Readme. Not sure what happened here

Mr-Leshiy:
@kickloop `cat-data-service` has not been merged into main branch, so actually you already updated where it is needed :)

PR #329, opened 2023-04-25
Cat data service proposals

cong-or:
[API](https://github.com/input-output-hk/catalyst-core/blob/a3cdbdc06fdee7368b3724eebced24b732f6f8b5/book/src/07_web_api/openapi/core_backend_api.yaml#L137)

p.s there is most likely conflicts with the other proposals endpoint currently under construction.


src/cat-data-service/src/service/v1/event/proposal.rs
```diff
@@ -1,7 +1,112 @@
...
+            handle_result(proposals_exec(path, query, state).await).await
+        }),
+    )
+}
+
+#[derive(Deserialize)]
+struct ProposalsQuery {
+    limit: Option<i64>,
+    offset: Option<i64>,
+    _voter_group: Option<VoterGroup>,
```
Mr-Leshiy, line 34: I think we should put TODO here

src/event-db/src/queries/event/proposal.rs
```diff
@@ -0,0 +1,102 @@
+use crate::{
+    error::Error,
+    types::event::{objective::ObjectiveId, proposal::ProposalSummary},
+    EventDB,
+};
+use async_trait::async_trait;
+use serde::Deserialize;
+
+#[derive(Deserialize)]
+pub enum VoterGroup {
```
Mr-Leshiy, line 10: Nit, it was discussed before, we don't have enums as a representation of some db fields, even if in our service API they are defined as enums. The main intention is not to have a strict validation of these fields in our service, so in case when we want to update or expand these values we will just need to update our API documentation and put new values into the db, and we don't want to update service.
So you should use just a wrapper over `String` type at this place
```
pub struct VoterGroup(pub String);
```
P.S. Here is an example the same field https://github.com/input-output-hk/catalyst-core/blob/0444b08ec66c80ae95cf8bbdbe606d0ed1bd6a06/src/event-db/src/types/event/mod.rs#L422. (if you will introduce a new type, can you pls modify this place too pls 😊)

src/event-db/src/types/event/proposal.rs
```diff
@@ -45,6 +45,19 @@ pub struct Proposal {
     pub proposal_details: ProposalDetails,
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct EventId(pub i32);
```
Mr-Leshiy, line 49: Duplication of this type https://github.com/input-output-hk/catalyst-core/blob/0444b08ec66c80ae95cf8bbdbe606d0ed1bd6a06/src/event-db/src/types/event/mod.rs#L10

src/event-db/src/types/event/proposal.rs
```diff
@@ -45,6 +45,19 @@ pub struct Proposal {
...
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct ObjectId(pub i32);
```
Mr-Leshiy, line 52: I think it should be moved into the `objective.rs` file

src/event-db/test_data/proposal_table.sql
```diff
@@ -0,0 +1,119 @@
+INSERT INTO
```
Mr-Leshiy, line 1: This query can be minimized in the following way
```
INSERT INTO proposal (....)
VALUES (.....), (....), (....);
```

Mr-Leshiy (changes requested):

Mr-Leshiy (approved):

PR #331, opened 2023-04-26
fix: Pad public key bytes if they are smaller than the expected size

FelipeRosa:


kukkok3 (commented):
Tested and I dont get the error anymore. Wait for a review from Steven anyway

stevenj (approved):
LGTM

PR #333, opened 2023-04-27
feat: use Earthly for deploying containers

saibatizoku:
Adds `Earthfile` for simpler deployment of the components in the monorepo.


Earthfile
```diff
@@ -0,0 +1,119 @@
...
+
+    WORKDIR /work
+
+    IF [ $EARTHLY_CI == "true" ]
+        ARG tag=$(TZ=UTC date +"%Y%m%d%H%M%S")-${EARTHLY_GIT_SHORT_HASH}
+    ELSE
+        ARG tag=latest
+    END
+
+    SAVE IMAGE builder:$tag
```
jmgilman, line 65: I don't think tagging it is necessary here. We should be able to just replace it with a cache hint.

Earthfile
```diff
@@ -0,0 +1,119 @@
...
+        chmod +x with_nix && \
+        sudo ln -s /devshell/with_nix /usr/bin/with_nix
+
+    WORKDIR /work
+
+    IF [ $EARTHLY_CI == "true" ]
+        ARG tag=$(TZ=UTC date +"%Y%m%d%H%M%S")-${EARTHLY_GIT_SHORT_HASH}
+    ELSE
+        ARG tag=latest
+    END
```
jmgilman, line 62: This should go away.

Earthfile
```diff
@@ -0,0 +1,119 @@
...
+    RUN sed -i 's/%sudo.*ALL/%sudo   ALL=(ALL:ALL) NOPASSWD:ALL/' /etc/sudoers && \
+        echo "sandbox = false" > /etc/nix.conf && \
+        echo "experimental-features = nix-command flakes" >> /etc/nix.conf
+
+    # Install Nix
+    USER $user
+    ENV USER=${USER}
+    ENV NIX_PATH=/home/${USER}/.nix-defexpr/channels:/nix/var/nix/profiles/per-user/root/channels
+    ENV NIX_CONF_DIR /etc
+    RUN curl -L 'https://nixos.org/nix/install' | NIX_INSTALLER_NO_MODIFY_PROFILE=1 sh
```
jmgilman, line 32: Probably should throw a cache hint here as well. The Nix install isn't super intensive, but downloading a cached result is certainly faster.

jmgilman (changes requested):

jmgilman (changes requested):
Looking good. Just need to port over the entrypoint scripts. 

Some notes on that:

1. Prefer to set the following bash options:

```
set +x
set -o errexit
set -o pipefail
set -o nounset
set -o functrace
set -o errtrace
set -o monitor
set -o posix
shopt -s dotglob
```

2. All entrypoints (that are bash scripts) should be contained in separate files. These files should be included in the CI process, namely they should be "dry run" (using `bash -n`) and put through `shellcheck` to detect common errors. 

3. All entrypoints should contain a `DEBUG_SLEEP` environment variable that, if present, causes the container to wait for the number of seconds assigned to the variable. For example, `DEBUG_SLEEP=3600` would cause the container to wait for 1 hour before proceeding. This is critical for troubleshooting in a live environment. 

4. All entrypoints should utilize `$@` when passing flags to the target binary. This way additional flags can be passed when the container is run.

stevenj (dismissed):
LGTM

minikin:
@saibatizoku Should we merge it?

jmgilman (dismissed):
LGTM.

minikin (approved):

jmgilman (approved):

PR #336, opened 2023-04-27
feat(cicd): Add starting point for migrations container build and CI.

stevenj:
Starting point for migrations container and an example of testing it in CI


saibatizoku (approved):
Thanks, LGTM.

PR #340, opened 2023-05-01
feat: Add option for reps file in catalyst-toolbox

FelipeRosa:
Added `--reps-file` option to catalyst-toolbox which allows a JSON file containing a list of voting keys to be used when deciding which group (direct or reps) to assign a voting key to.

Also made the code for deserializing voting keys public in `snapshot_lib`.


FelipeRosa:
Closing in favor of #341

PR #342, opened 2023-05-03
feat: Implement `api/v1/event/{id}/{ref_id}/{prop_id}/review_types` endpoint

Mr-Leshiy:
# Description

Add a new `api/v1/event/{id}/{ref_id}/{prop_id}/review_types` endpoint with the db types and queries.


stevenj (approved):
LGTM

PR #349, opened 2023-05-11
refactor: Makes migrations more configurable

jmgilman:
Improves the entrypoint script for the migrations container by:

- Adding common bash options
- Adding support for debugging and sleeping
- Documenting the expected environment variables
- Allowing skipping steps of initialization (creating the database/user is not required in Kubernetes)
- Inheriting from base builder/deployment images to be consistent with other containers
- Running commands as expected database user

Additionally, this PR modifies the SQL setup script to stop printing passwords to `stdout`. Note that container logs will reflect anything in `stdout` and are, by default, exported to Loki. We should, therefore, avoid printing any sensitive values to `stdout`. You can still set the `DEBUG` environment variable and the passwords will show up in the printed SQL commands.

Finally, the image is bigger due to using debian instead of alpine. This is the cost of consistency.

### DIFF
--- a/containers/event-db-migrations/entry.sh
+++ b/containers/event-db-migrations/entry.sh
@@ -1,46 +1,147 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
 
-# Set default value for reinitDb flag
-reinitDb=false
+# ---------------------------------------------------------------
+# Entrypoint script for migrations container
+# ---------------------------------------------------------------
+#
+# This script serves as the entrypoint for the migrations container. It sets up
+# the environment, performing optional database initialization if configured,
+# and then runs the migrations.
+#
+# It expects the following environment variables to be set except where noted:
+#
+# DB_HOST - The hostname of the database server
+# DB_PORT - The port of the database server
+# DB_NAME - The name of the database
+# DB_SUPERUSER - The username of the database superuser
+# DB_SUPERUSER_PASSWORD - The password of the database superuser
+# DB_USER - The username of the database user
+# DB_USER_PASSWORD - The password of the database user
+# ADMIN_ROLE_PASSWORD - The password of the cat_admin role for graphql
+# ADMIN_USER_PASSWORD - The password of the admin user for graphql
+# ANON_ROLE_PASSWORD - The password of the cat_anon role for graphql
+# ADMIN_FIRST_NAME - The first name of the admin user for graphql (optional)
+# ADMIN_LAST_NAME - The last name of the admin user for graphql (optional)
+# ADMIN_ABOUT - The about of the admin user for graphql (optional)
+# ADMIN_EMAIL - The email of the admin user for graphql (optional)
+# REINIT_EVENT_DB - If set, the database will be reinitialized (optional) (DESTRUCTIVE)
+# SKIP_EVENT_DB_INIT - If set, the event database will not be initialized (optional)
+# SKIP_GRAPHQL_INIT - If set, graphql will not be initialized (optional)
+# DEBUG - If set, the script will print debug information (optional)
+# DEBUG_SLEEP - If set, the script will sleep for the specified number of seconds (optional)
+# ---------------------------------------------------------------
 
-# Check if reinitDb flag is set in environment variable
-if [ -n "$REINIT_DB" ]; then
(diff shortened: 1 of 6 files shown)


containers/event-db-migrations/entry.sh
```diff
@@ -1,46 +1,147 @@
...
+
+: "${ADMIN_FIRST_NAME:='Admin'}"
+: "${ADMIN_LAST_NAME:='Default'}"
+: "${ADMIN_ABOUT:='Default Admin User'}"
+: "${ADMIN_EMAIL:='admin.default@projectcatalyst.io'}"
+
+# Sleep if DEBUG_SLEEP is set
+debug_sleep
+
+# Initialize database if necessary
```
stevenj, line 95: I don't know why we need `/tmp/initialize` at all?

We should be able to run the graphql setup multiple times without concern.
The ONLY issue with the graphql setup is deleting the user table.

I propose we just have the env vars which control initing the base DB and graphql ONLY.  
I also think it shouldn't be a SKIP env var it should be a "INIT_EVENT_DB" and "INIT_GRAPHQL" env var,  default to false.

src/event-db/setup/graphql-setup.sql
```diff
@@ -60,13 +60,10 @@
 \echo VARIABLES:
 \echo -> dbName ....................... = :dbName
 \echo -> dbUser ....................... = :dbUser
-\echo -> adminRolePw / $ADMIN_ROLE_PW . = :adminRolePw
-\echo -> anonRolePw / $ANON_ROLE_PW ... = :anonRolePw
 \echo -> adminUserFirstName ........... = :adminUserFirstName
 \echo -> adminUserLastName ............ = :adminUserLastName
 \echo -> adminUserAbout ............... = :adminUserAbout
 \echo -> adminUserEmail ............... = :adminUserEmail
-\echo -> adminUserPw / $ADMIN_USER_PW . = :adminUserPw
```
stevenj, line 67: I know your deleting these for security.  Can we instead just put `***` as the value.
The reason being that I like this publishing all possible options from a self documentation perspective.

stevenj (changes requested):

containers/event-db-migrations/entry.sh
jmgilman: The container will be configured as a Kubernetes Job that will be deployed with the rest of the applications. Since it's a job, it will only run once. However, there is a small chance of it inadvertently being run again (due to operator error), and if it's been configured with `INIT_EVENT_DB`, then the event database gets wiped. Whereas, assuming the volume is mounted properly, with the `/tmp/initialize` approach the script would refuse to initialize the database unless the `REINIT_EVENT_DB` environment variable was given. 

So, by default we initialize, unless we already know it's initialized. I'm not a fan of using a volume for tracking state. If you know of a better way to determine when we should initialize, I would prefer a better approach.

jmgilman: Yeah, that's sort of what I was trying to get at in Slack. We would need something to validate that both initializations ran. For example, maybe for the `event_db` initialization we can verify the expected user/database exists. For the graphql stuff, maybe just look for the `private` schema?

@stevenj Thoughts?

saibatizoku (dismissed):
LGTM

containers/event-db-migrations/entry.sh
stevenj: Yes, something exactly like this.

stevenj: For the DB the very last thing we could do when initialising it is write a record to the `config` table eg:
```
`id`: "Initialized", id2: "", id3: "", value: {"at" : "<datetime>"}
```

Because this is the VERY last thing we do when we init the database, we know its been done if that record exists.
If it stops at any point before that the DB is screwed and needs to be re-initialized anyway.

We don;t need to check the value of the `value` just that we can select this record in the `config` table and it returns data.

stevenj: Because graphql init is needed,  we don't really need a separate check for it.  IF we have this record graphql was setup along with everything else.

jmgilman: We can tackle this in another PR.

saibatizoku (approved):
LGTM.

PR #350, opened 2023-05-11
feat(ci): Create stale-branches.yml

minikin:
Add stale-branches.yml job


Mr-Leshiy (approved):
I think we need the same on the `neo` repository

PR #354, opened 2023-05-15
fix: Catalyst-toolbox sve-snapshot command logic

FelipeRosa:


kukkok3 (changes requested):
Tested and it looks good. Could you also add 
```
{
  "initial":[
  {
    "fund":
```
at the beginning of the snapshot when generated as json?

kukkok3 (changes requested):
Outuput is:
```
{
  "initial": {
    "fund":
```    
but is should be
```
 {
  "initial": [ 
  {
    "fund":
```
  missing a **[** after` "initial":`

stevenj (approved):

kukkok3 (approved):

PR #376, opened 2023-05-16
feat: Add event-db eathly tests running

Mr-Leshiy:
# Description

Fixed Earthly files to be functional on the macos M1.
Added event-db earhtly tests running target


containers/event-db-migrations/entry.sh
```diff
@@ -40,7 +40,21 @@ fi
...
   psql -U $PGUSER -d $PGDATABASE -h $PGHOST -p $PGPORT -f "$file"
 done
+fi
+
+# Add test data
+directory="./test_data"
+
+if [ -d "$directory" ];
+then
+for file in $(cd "$directory" && ls *.sql | sort -n); do
```
jmgilman, line 56: It's best practice to use `find` in these scenarios. See an example here: https://github.com/input-output-hk/catalyst-core/pull/349/files#diff-fe4e0fde017cc32263b0093d0e53b072b171d4567f61332d7dcc89230c1d2d57R142

jmgilman (changes requested):

Mr-Leshiy:
Need to merge this firstly https://github.com/input-output-hk/catalyst-core/pull/349

containers/event-db-migrations/Earthfile
```diff
@@ -40,6 +40,33 @@ docker:
     # Push the container...
     SAVE IMAGE --push ${registry}migrations:$tag
 
+docker-test:
```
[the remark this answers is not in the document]
Mr-Leshiy: I see, lets merge your PR first and then I will update mine.

Earthfile
```diff
@@ -2,66 +2,9 @@
...
-    RUN rm rustup-init rustup-init.sha256
-
-    # Force rustup to initialize the toolchain from the rust-toolchain file
-    COPY rust-toolchain .
-    RUN rustup show
-
 # Installs Cargo chef
 install-chef:
-    FROM +rust-toolchain
+    FROM rust:latest
```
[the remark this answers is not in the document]
Mr-Leshiy: I have changed it to the multi-platform image, to prevent specifying `--platform` flag, and to use a native `environment` if it is possible (at least as I understood how it works). 
Another thing is that when it was a `debian:stable-slim`, on my local apple M1 machine it was a problem with the earthly buildkit grpc connection which I have mentioned in slack.

containers/event-db-migrations/entry.sh
Mr-Leshiy: fixed

Earthfile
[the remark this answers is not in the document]
jmgilman: The [debian:stable-slim image](https://hub.docker.com/layers/library/debian/stable-slim/images/sha256-1529cbfd67815df9c001ed90a1d8fe2d91ef27fcaa5b87f549907202044465cb?context=explore) is built against all major platforms. You can test it yourself by doing:

```
docker run -it --entrypoint /bin/bash debian:stable-slim
```

If you get into a shell, then it's a valid multi-platform image and isn't the root cause of your issues. For the GRPC issue, did you try wiping the buildkit container and doing `earthly --bootstrap` again? I can't imagine the issue being related to a specific image, especially if the command I gave above works.

The only reason I'm against changing to `rust:latest` is because if we build against one image and deploy on another, there's a huge chance for things breaking. We need to standardize on a single image for both building/deployment, and I would imagine `rust:latest` ships with a whole bunch of dependencies we don't want in production.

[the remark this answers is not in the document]
Mr-Leshiy: Updated and it works for me !

Mr-Leshiy:
@jmgilman ready for review

containers/event-db-migrations/entry.sh
```diff
@@ -133,15 +132,18 @@ fi
 # Run migrations
 echo ">>> Running migrations..."
 export DATABASE_URL="postgres://${DB_USER}:${DB_USER_PASSWORD}@${DB_HOST}:${DB_PORT}/${DB_NAME}"
+echo "$DATABASE_URL"
```
jmgilman, line 135: Leftover from debugging?

containers/event-db-migrations/entry.sh
```diff
@@ -133,15 +132,18 @@ fi
...
 
 # Add historic data from previous funds
-export PGUSER="${DB_USER}"
-export PGPASSWORD="${DB_USER_PASSWORD}"
-
 while IFS= read -r -d '' file; do
     echo "Adding historic data from $file"
     psql -f "$file"
 done < <(find ./historic_data -name '*.sql' -print0 | sort -z)
+# Add test data
```
jmgilman, line 143: Have you tried testing this if `./test_data` doesn't exist? Since the `-e` opt is set at the top, if `find` returns a non-zero exit code it may terminate the script early.

src/cat-data-service/src/service/v1/event/mod.rs
```diff
@@ -61,7 +61,15 @@ async fn events_exec(
 }
 
 /// Need to setup and run a test event db instance
-/// To do it you can use `cargo make local-event-db-test`
+/// To do it you can use the following commands:
```
jmgilman, line 64: We could probably encapsulate all of these steps into a single Bash script.

jmgilman (changes requested):

containers/event-db-migrations/entry.sh
Mr-Leshiy: yes, it works in both sides, if test_data or historic_data does not exist

src/cat-data-service/src/service/v1/event/mod.rs
Mr-Leshiy: I think it will be redundant, it is needed only if you want to run it locally

containers/event-db-migrations/Earthfile
```diff
@@ -46,7 +40,12 @@ docker:
     # Add migrations
     COPY --dir ../../src/event-db+build/setup ./setup
     COPY --dir ../../src/event-db+build/migrations ./migrations
-    COPY --dir ../../src/event-db+build/historic_data ./historic_data
+    IF [ "$data" = "historic" ]
+        COPY --dir ../../src/event-db+build/historic_data ./historic_data
+    END
```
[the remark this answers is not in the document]
Mr-Leshiy: fixed

containers/event-db-migrations/entry.sh
Mr-Leshiy: fixed

jmgilman (approved):
Assuming  you've run the `+all` target and everything is building, this LGTM.

PR #396, opened 2023-05-18
feat: NPH-6380 Update supplemental fields

Mr-Leshiy:
# Description

Updated ProposalSupplementalDetails and ObjectiveSupplementalData types to be a wrapper over a json object, without any fixed fields.


cong-or (approved):
LGTM

PR #404, opened 2023-05-24
feat: Update vote plan schema | NPG-6404

Mr-Leshiy:
# Description

Updated a vote plan schema
- modified `vote_plan.id` column, allow to be a non unique
- moved `token_id` column from `voting_group` table to `vote_plan` table
- updated `voting_group` table, left only only one column `name`
- modified `vote_plan` table, replaced reference to `event` table with the `objective` table (`event_id` column replaced with the `objective_id` column)
- updated `Event` type, removed `groups` field


saibatizoku (approved):
LGTM.

PR #418, opened 2023-06-01
refactor: Cat-data-service tests update | NPG-6632

Mr-Leshiy:
# Description

Refactored cat-data-service tests, updated assertions checks with the json representation of the returning object without using explicit Rust db types. It allows to test precisely by our API specification, instead of using our Rust internal db types serde.
Also updated `map` field of the `ReviewType` from string to object.
Updated `review_metric` chema, `map` column from `JSONB` to `JSONB ARRAY`

### DIFF
--- a/book/src/07_web_api/openapi/core_backend_api.yaml
+++ b/book/src/07_web_api/openapi/core_backend_api.yaml
@@ -1100,7 +1100,6 @@ components:
             * True - **MUST** include a note.
         map:
           type: array
-          item: object
           description: |-
             Optional sequential list of mapped named values for rating scores.
             * If not present, the rating score is numeric.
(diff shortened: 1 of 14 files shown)


cong-or (approved):
LGTM

book/src/07_web_api/openapi/core_backend_api.yaml
```diff
@@ -1100,7 +1100,7 @@ components:
             * True - **MUST** include a note.
         map:
           type: array
-          item: string
+          item: object
```
stevenj, line 1103: If its being suggested it be `{"name":<name>, "desc":<desc>}` then this should be defined here.
`object` doesn't tell an API consumer what they should expect.

Also, the `example` is now wrong as well.

src/cat-data-service/src/service/v1/event/objective/review_type.rs
```diff
@@ -77,51 +77,51 @@ mod tests {
...
+                        "group": "review_group 2"
+                    },
+                    {
+                        "id": 5,
+                        "name": "vpa_ranking",
+                        "description": "VPA Ranking of the review",
+                        "min": 0,
+                        "max": 3,
+                        "map": [
+                            {"name": "Excellent", "desc": "Excellent Review"},
```
stevenj, line 116: If this is what your proposing, then the API should explicitly state its "name" and "desc" what they mean and if they are optional. ("desc" should be optional, "name" not).

stevenj (changes requested):

src/cat-data-service/src/service/v1/event/objective/review_type.rs
Mr-Leshiy: Yes, will add a specific types and descriptions into the API doc, but will not modify db schema and how I am processing its, will let it as general as possible (array of json objects).

stevenj (approved):
LGTM

PR #423, opened 2023-06-06
fix: Even-db queries, objective id column | NPG-6632

Mr-Leshiy:
# Description

Fixed some queries of the event-db. Issue was with `WHERE` conditions which were relied on the `objective.raw_id` column as a provided by the user information (which actually should be hidden as an internal representation of the db). Replaced it to the `objective.id` comlumn.


cong-or (approved):
LGTM

saibatizoku (approved):

PR #424, opened 2023-06-06
fix: Fix pin type | NPG-6509

kukkok3:
Fixing a bug in this cat-toolbox cmd
```
catalyst-toolbox qr-code info --input wallet_2_above_500_1234.png --pin 1234 --opts img
The application panicked (crashed).
Message:  Mismatch between definition and access of pin. Could not downcast to TypeId { t: 3006200894326385576 }, need to downcast to TypeId { t: 4144576253896540664 }
```


Mr-Leshiy (approved):
LGTM

PR #425, opened 2023-06-07
fix: Remove sve snapshot python script | NPG-000

kukkok3:
Removing sve snapshot python script because the cmd has been introduced into catalyst-toolbox


stevenj (approved):
LGTM

PR #427, opened 2023-06-08
fix: Fix get_proposal query, updated responded error message | NPG-6702

Mr-Leshiy:
# Description

- Fixed `get_proposal` query, slightly updated tests
- Updated responded error message from the cat-data-service, so now instead of returning just a string, it  returns a `{error: "some error" }` json object.


cong-or (approved):
LGTM

stevenj (approved):
LGTM.
Great turnaround on this.

PR #429, opened 2023-06-12
feat: Add vit-ss compatible fund query | NPG-6690

Mr-Leshiy:
# Description

Added a vit-ss backward compatible `fund` query. This query retriewes an information about the current fund with the following query logic:
- `select ... from event where  event.end_time > CURRENT_TIMESTAMP and event.start_time < CURRENT_TIMESTAMP order by event.row_id desc limit 1`. So we are acutlly looking to the currently running fund, and for the case when we will have multiple running events at the same time, we will return info only about one with the highest id.


cong-or (dismissed):
lgtm

src/event-db/src/types/vit_ss/vote_plan.rs
```diff
@@ -0,0 +1,59 @@
+use crate::types::utils::serialize_datetime_as_rfc3339;
```
stevenj, line 1: Are you using the views defined in `V9__vitss_compatibility.sql`??

stevenj (commented):
See comments

src/event-db/src/types/vit_ss/vote_plan.rs
Mr-Leshiy: No, I have not relied on `V9__vitss_compatibility.sql`, I have added brand new queries. Of course they are mostly the same what we can see inside `V9__vitss_compatibility.sql`, but it is a separate queries.

stevenj (approved):
LGTM

PR #438, opened 2023-06-20
feat: Add script to extract F9 voteplans and generate SQL statements | NPG-6851

FelipeRosa:


minikin:
@FelipeRosa @saibatizoku @stevenj Shouldn't we add ci checks for Python?

stevenj (approved):
LGTM

PR #447, opened 2023-06-23
chore: Adjust voting node service to work with the merged snapshot config | NPG-6874

FelipeRosa:
I also changed the version in the importer's pyproject.toml.


stevenj (approved):
LGTM

PR #449, opened 2023-06-27
feat: Added cat-data-service-client | NPG-6476

nicopado:


stevenj:
Is all this code (or the majority) autogenerated?

stevenj:
> Yes it is all autogenerated from the openapi spec with https://openapi-generator.tech/docs/generators/rust/ I did a couple of manual correction to it, nothing major

In general its not a good idea to put autogenerated code/data into git, because it will/should change every time we update the ap[I yaml file.

It would be better if this was just the scripting to run the generator,  and then .gitignore files for the resulting generated data.
So that it can be re-generated on-damand,  not a potentially out of date version being pulled from git.

nicopado:
> > Yes it is all autogenerated from the openapi spec with https://openapi-generator.tech/docs/generators/rust/ I did a couple of manual correction to it, nothing major
> 
> In general its not a good idea to put autogenerated code/data into git, because it will/should change every time we update the ap[I yaml file.
> 
> It would be better if this was just the scripting to run the generator, and then .gitignore files for the resulting generated data. So that it can be re-generated on-damand, not a potentially out of date version being pulled from git.

I absolutely agree on that, in general.
In the specific case tho it might be a little bit more work intensive to fully automate it as there also are some manual changes to do to the autogen code in order to make it work. Nothing that can't be automated with scripting of course, but I think that this chunk of automation can be delayed so that we can proceed with the testing quicker and leave this automation for the future.
If you think it's best to automate it right now I can delay the tests in favour of this, your call :)

minikin:
@nicopado 
> there also are some manual changes to do to the autogen code in order to make it work

What exactly do we need to change in autogenerated code? That sounds like we probably doing something wrong.

minikin:
@nicopado should we close this one?

PR #455, opened 2023-07-03
refactor: Move serde definiton from event-db to cat-data-service (Part 1) | NPG-7002

Mr-Leshiy:
# Description

Moving serde implementation of the types which are cunsumed by the cat-data-service directly into the cat-data-service from event-db. 
It makes cat-data-service independent from the event-db of how represent the data while we are returning it in responses.


stevenj (approved):
LGTM

PR #457, opened 2023-07-04
refactor: Move serde definiton from event-db to cat-data-service (Part 2) | NPG-7002

Mr-Leshiy:
Moving serde implementation of the types which are cunsumed by the cat-data-service directly into the cat-data-service from event-db.
It makes cat-data-service independent from the event-db of how represent the data while we are returning it in responses.


src/cat-data-service/src/types/proposal.rs
```diff
@@ -0,0 +1,281 @@
...
+        SerdeType(&self.0).serialize(serializer)
+    }
+}
+
+impl Serialize for SerdeType<&ProposerDetails> {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let mut serializer = serializer.serialize_struct("ProposerDetails", 4)?;
```
minikin, line 65: Do we have any options not to use dynamic dispatch? We know the shape of the data in compilation time, right?

src/cat-data-service/src/types/proposal.rs
```diff
@@ -0,0 +1,281 @@
...
+    }
+
+    #[test]
+    fn proposer_details_json_test() {
+        let proposer_details = SerdeType(ProposerDetails {
+            name: "proposer name".to_string(),
+            email: "proposer email".to_string(),
+            url: "proposer url".to_string(),
+            payment_key: "proposer payment key".to_string(),
+        });
```
minikin, line 186: White spaces scare me. Does this data come from CSV (importer) or something?

minikin (commented):
I left two comments, thanks!

src/cat-data-service/src/types/proposal.rs
Mr-Leshiy: Not sure that it is possible, because it is the only way how to define serializations for structs.
Also if deeply take a look under the `derive(Serialize)` implementation of serde we will the same approach
https://github.com/serde-rs/serde/blob/992a01bad2c65c2920296cdb0d6708bdd78eae4c/serde_derive/src/ser.rs#L350C1-L351C1

Mr-Leshiy: Or what do you mean here ?

minikin: Discussed offline.

minikin (approved):

PR #485, opened 2023-07-25
feat: allow skipping insertion of historical/test data in migrations …

jmgilman:
Currently, this is being controlled when the container is built. If a specific flag is passed, the data will either be present or absent when the container is built, thus causing the for loop to execute or not. This isn't really ideal, since we prefer to have the container ship with everything and then configure what gets inserted at runtime. 

I've added environment variables to control this at runtime. I have not modified the Earthfile to change the previous logic because I'm not sure if it's actively being used anywhere.


stevenj (approved):
LGTM

PR #490, opened 2023-07-27
feat: Audit tooling | NPG-6520

cong-or:
More granular docs will follow in another PR.

Keep in my mind, code from _recover.rs_ has been **salvaged** from other parts of the code base...


stevenj (approved):
LGTM

minikin:
@cong-or 

```sh
error: failed to compile `cargo-make v0.36.12`, intermediate artifacts can be found at `/tmp/cargo-installsZuhxD`

Caused by:
  package `time v0.3.24` cannot be built because it requires rustc 1.67.0 or newer, while the currently active rustc version is 1.65.0
  Try re-running cargo install with `--locked`
Error: Process completed with exit code 101.
```

Should we consider updating to the latest stable `1.71.0` ?

PR #491, opened 2023-07-28
feat: Update catalyst-toolbox `proposal-score` command | NPG-7776

Mr-Leshiy:
# Description

Modified `proposal-score` cli command, so now it is not inserted proposal's scores directly into the database.
It inserts data into the provided `proposals.json` file.
This update will simplify deployment process so the flow would the following:
1. Run ideascale importer to get `proposals.json` file
2. Take `reviews.csv` file with all reviews
3. Run catalyst-toolbox `proposal-score` command and provide `reviews.csv` as an input, `proposals.json` as an input and as an output of the updated proposals with the added scores data


src/catalyst-toolbox/catalyst-toolbox/src/proposal_score/store.rs
```diff
@@ -1,75 +1,97 @@
-use super::{AligmentScore, AuditabilityScore, FeasibilityScore, ProposalId};
-use rusqlite::Connection;
-use std::{path::PathBuf, str::FromStr};
+use super::{AligmentScore, AuditabilityScore, FeasibilityScore};
```
minikin, line 1: @Mr-Leshiy typos in `Aligment***` types, should be `Alignment***`

src/catalyst-toolbox/catalyst-toolbox/src/proposal_score/load.rs
```diff
@@ -67,16 +73,57 @@ pub fn load_reviews_from_csv(
     Ok(reviews_per_proposal)
 }
 
+pub fn load_proposals_from_json(
```
minikin, line 76: I was wondering if it would be reasonable to move the parsing proposal ID operation into its own function? 
Moreover, we probably can consider using `map` instead of `loop` if we prepare a functional approach.

```suggestion
pub fn load_proposals_from_json(
    path: &Path,
) -> Result<HashMap<ProposalId, serde_json::Value>, Error> {
    let proposals: Vec<serde_json::Value> = serde_json::from_str(&std::fs::read_to_string(path)?)?;
    let mut res = HashMap::new();
    let res: Result<HashMap<_, _>, _> = proposals
        .into_iter()
        .map(|proposal| Ok((parse_proposal_id(&proposal)?, proposal)))
        .collect();
    Ok(res)
}

fn parse_proposal_id(proposal: &serde_json::Value) -> Result<ProposalId, Error> {
    Ok(ProposalId(
        proposal
            .get("proposal_id")
            .ok_or_else(|| {
                Error::InvalidProposalData("does not have \"proposal_id\" field".to_string())
            })?
            .as_str()
            .ok_or_else(|| {
                Error::InvalidProposalData(
                    "invalid \"proposal_id\" field type, not a string".to_string(),
                )
            })?
            .parse()
            .map_err(|_| {
                Error::InvalidProposalData(
                    "invalid \"proposal_id\" field type, not a string encoded int".to_string(),
                )
            })?,
    ))
}
```

minikin (changes requested):

src/catalyst-toolbox/catalyst-toolbox/src/proposal_score/load.rs
minikin: looks like the formating of the suggestion is off...

minikin: Or maybe we can do something like this:

```rust
pub fn load_proposals_from_json(
    path: &Path,
) -> Result<HashMap<ProposalId, serde_json::Value>, Box<dyn Error>> {
    let file = File::open(path)?;
// Reading the entire file into memory before parsing it can be inefficient, especially for large JSON files. 
// Instead,  we can read the file in chunks using a buffered reader.
    let reader = BufReader::new(file);

    let mut res = HashMap::new();
// Instead of reading the entire JSON file as a string and then parsing it, 
// we can use the `serde_json::Deserializer` to directly parse the file into Rust objects. 
// This avoids the intermediate step of creating a string representation of the JSON data.
    let mut de = serde_json::Deserializer::from_reader(reader).into_iter::<Value>();

    while let Some(proposal) = de.next() {
        let proposal = proposal?;
        let proposal_id = parse_proposal_id(&proposal)?;
        res.insert(proposal_id, proposal);
    }

    Ok(res)
}
```

src/catalyst-toolbox/catalyst-toolbox/src/proposal_score/store.rs
Mr-Leshiy: done

src/catalyst-toolbox/catalyst-toolbox/src/proposal_score/load.rs
Mr-Leshiy: done

minikin (approved):

PR #499, opened 2023-08-01
chore: Correct event-db schema version | NPG-7769

FelipeRosa:
# Description

Change event-db schema version to 9 so it works after the removal of the vit-ss compatibility tables.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)


jmgilman (approved):
LGTM.

PR #507, opened 2023-08-10
feat: Remove deprecated fields | NPG-0000

Mr-Leshiy:
# Description

Removed vit-ss deprecated columns from event-db tables


stevenj (dismissed):
Actually,  While I would like to just remove from the base schema,  we need to migrate now.
So this change actually needs to be a V10 migration file which deletes these columns.

stevenj (changes requested):
Don't delete here, add a new migration which drops the columns.

minikin:
@Mr-Leshiy, What do you think we should do about this PR?

PR #510, opened 2023-08-11
feat: Add initial version of the reviews importer | NPG-7896

Mr-Leshiy:
# Description

Added new reviews importer.
It is an initial version and non fully complete, right now it is able only dowload xls files from the ideascale.
Later it is needed to process dowloaded data and insert it into db


utilities/ideascale-importer/ideascale_importer/reviews_importer.py
```diff
@@ -0,0 +1,102 @@
+from lxml import html
+import uuid
+
+from ideascale_importer import utils
+
+def random_string(string_length=6):
```
saibatizoku, line 6: A suggestion to do the same in one line:

```
import secrets

....
def random_string(string_length=6):
    return secrets.token_hex(string_length)
```

saibatizoku (approved):
LGTM

utilities/ideascale-importer/ideascale_importer/reviews_importer.py
```diff
@@ -0,0 +1,96 @@
...
+            item  = int(tree.find(".//div[@data-processing-item-id]").get("data-processing-item-id"))
+
+            export_data_endpoint = "/a/reporting/export-data/"
+            while True:
+                content = await self.inner.get(f"{export_data_endpoint}{item}")
+                if "Finished Processing" in str(content):
+                    download_endpoint = "/a/download-export-file/"
+
+                    content = await self.inner.get(f"{download_endpoint}{item}")
+                    f = open(f"{out_dir}/{file_name}.xls", "wb")
```
stevenj, line 48: Instead of using temporary files, it would be better to actually read the excel data here as its read, and convert it.
You can use https://openpyxl.readthedocs.io/en/stable/api/openpyxl.reader.excel.html to read excel files.
And `io.BytesIO(content)` to convert the content downloaded into an in-memory file like object which the openpyxl reader can handle.
This will also tell you right here if there is an issue with the data being downloaded or not.

stevenj (changes requested):

utilities/ideascale-importer/ideascale_importer/reviews_importer.py
Mr-Leshiy: done

stevenj (approved):
LGTM

PR #518, opened 2023-08-15
feat: Move review's processing code from catalyst-export repo | NPG-7896

Mr-Leshiy:
# Description

Move review's processing code from catalyst-export repo. 
It is just a simple copy-past of the python modules, no actuall processing at this moment has been added into import implementation.
Initially I have tried to smartly move this code to avoid duplication of some parts, but unfortunatelly it does not seems achievable at least in reasonable amount of time.


FelipeRosa (approved):
LGTM

PR #521, opened 2023-08-16
feat: Add allocation generation step and preparing reviews for import | NPG-7896

Mr-Leshiy:
# Description

- Added allocations generation step to review importer tool.
- Added preparing reviews step to review importer tool.
- Update config loading, load configs from the event-db.


utilities/ideascale-importer/ideascale_importer/reviews_importer/processing/prepare.py
```diff
@@ -0,0 +1,82 @@
...
+        await ideascale.import_proposals(stage_ids=stage_ids)
+        await ideascale.import_com_revs(group_id=group_id, start_id=anonymize_start_id, historic_pas=importer.pas)
+        allocator.allocate()
+
+        # This data is not used by our core system, it is only consumed by Ideascale
+
+        # allocator.export_allocations(allocator.source.pas, f"{output_path}/allocations-by-pa.xlsx")
+        # groups = allocator.generate_challenges_groups()
+        # for group in groups:
+        #     allocator.export_allocations(group["pas"], f"{output_path}/{group['challenge']}.xlsx")
```
FelipeRosa, line 42: Why is this commented out?

FelipeRosa (approved):
LGTM

utilities/ideascale-importer/ideascale_importer/reviews_importer/processing/prepare.py
Mr-Leshiy: Right now we are not exporting data in this format into the excel files, but probably we will want to automate this part also. But right now it is not in current scope, so that is why I left it as commented out

stevenj (approved):
LGTM

Issue #523, opened 2023-08-18
earthly +local fails


profd2004:
**Describe the bug**
earthly +local local command fails.

**Steps to reproduce the behavior:**

1. Clear earthly cash
2. run `earthly +local`
<img width="250" alt="Screenshot 2023-08-17 at 9 58 33 PM" src="https://github.com/input-output-hk/catalyst-core/assets/4429661/129aa5ec-294f-4e2e-b841-5bb6d694adb1">   

**Expected behavior**
earthly +local should complete successfully with out additional host system dependencies outside of docker

**Desktop (please complete the following information):**
Mac OS 14.0
Docker for mac 4.21.1 (114176)

### Additional contexts:

`
./services/voting-node+build |   • Installing watchfiles (0.19.0)
./services/voting-node+build |   • Installing websockets (11.0.3)
        +build-cache |    Compiling multiparty v0.1.0
        +build-cache |    Compiling async-stream-impl v0.3.5
        +build-cache |    Compiling time v0.1.45
        +build-cache |    Compiling scoped-tls v1.0.1
                          Compiling iana-time-zone v0.1.56
        +build-cache |
./services/voting-node+build |   ChefBuildError

./services/voting-node+build |   Backend subprocess exited when trying to invoke get_requires_for_build_wheel
./services/voting-node+build |
./services/voting-node+build |   Building lxml version 4.9.0.
./services/voting-node+build |   <string>:67: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
./services/voting-node+build |   Building without Cython.
./services/voting-node+build |   Error: Please make sure the libxml2 and libxslt development packages are installed.
./services/voting-node+build |

./services/voting-node+build |   at ~/.local/share/pypoetry/venv/lib/python3.11/site-packages/poetry/installation/chef.py:147 in _prepare
./services/voting-node+build |       143│
./services/voting-node+build |       144│                 error = ChefBuildError("\n\n".join(message_parts))
./services/voting-node+build |       145│
./services/voting-node+build |       146│             if error is not None:
./services/voting-node+build |     → 147│                 raise error from None
./services/voting-node+build |       148│
./services/voting-node+build |       149│             return path
./services/voting-node+build |       150│
./services/voting-node+build |       151│     def _prepare_sdist(self, archive: Path, destination: Path | None = None) -> Path:

[...]

PR #525, opened 2023-08-18
build: adds missing build-time deps to voting-node | NPG-0000

jmgilman:
This fixes the current issue on #523. However, I do wonder if `build-essential` is actually required in the final container?


minikin (approved):
LGTM

PR #526, opened 2023-08-22
feat: Add option to specify the voting node's role | NPG-0000

FelipeRosa:
This PR adds the option to specify the voting node's role (instead of only parsing it from hostname). This is to allow for the voting node to work as, for example, leader0 and use a different hostname/ip address in its configuration.


saibatizoku:
Looks good. The new env var should probably be included in `entry.sh` as well.

stevenj (approved):
LGTM

PR #534, opened 2023-08-25
fix(catalyst-toolbox): Fix error when purpose is null. | NPG-0000

stevenj:
Catalyst toolbox was crashing if "purpose" was Null in a registration.
This fixes that, so Null is properly detected as a "Catalyst" registration.

Also, it adds the pre-calculated Jormungandr chain address,  because its non trivial to go from the voting key to the chain address used by Jormungandr.

This lets me compare the SVE snapshot with the normal snapshot, and it shows the SVE snapshot has rounding errors which will negatively effect people with multiple registrations on the same key.

Recommendation is for Audit, we publish the updated `snapshot` artifact, not the one we import into vit-ss.  And then make a quick program to convert the `snapshot` artifact into a format needed for vit-ss for F10.  The `utilities/snapshot-check/compare_snapshot.py` would need a couple of tweaks to be such a tool.


kukkok3 (dismissed):
LGTM

src/vit-testing/mainnet-tools/src/snapshot/convert.rs
```diff
@@ -130,7 +130,7 @@ impl OutputExtension for SnapshotEntry {
                     VotingDelegations::New(new)
                 }
             },
-            voting_purpose: *self.voting_purpose.unwrap_or(VotingPurpose::CATALYST),
+            voting_purpose: Some(*self.voting_purpose.unwrap_or(VotingPurpose::CATALYST)),
```
minikin, line 133: Can we remove `unwrap_or` instead of wrapping `*self.voting_purpose` in `Some`?

minikin (dismissed):

src/vit-testing/mainnet-tools/src/snapshot/convert.rs
stevenj: voting_purpose is an Option.  We MAY be able to remove the unwrap, but it's not clear to me what this test is even doing, so this keeps the logic as close as possible to the original, with the hope to preserve the tests functionality (whatever that is).

PR #536, opened 2023-08-25
fix: Update lxml for voting node | NPG-000

kukkok3:


saibatizoku (approved):
LGTM. Builds locally without a glitch.

PR #538, opened 2023-08-28
feat: Feature/poem migration start | NPG-0000

stevenj:
# Description

This is the start of the integration of Poem as the API server.
It is done in a way to allow the existing and new API server to co-exists on different ports.
This can be maintained until all endpoints have been ported.

This is general code architecture work.

It also includes a start of a local monitoring stack, which will be improved upon in a future PR.


Mr-Leshiy (commented):
Am I correct that we will move all our docker-compose files under the `containers/dev-local` dir and so these docker-compose files should be used also for testing purposes ?

services/voting-node/poetry.toml
```diff
@@ -4,7 +4,7 @@ prefer-active-python = true
 
 [virtualenvs.options]
 no-pip = true
-no-setuptools = true
+#no-setuptools = true
```
Mr-Leshiy, line 7: Is it intentional ?

src/cat-data-service/src/service/utilities/metrics_tracing.rs
```diff
@@ -0,0 +1,39 @@
...
+use poem::{middleware::{CorsEndpoint, OpenTelemetryMetricsEndpoint}, Endpoint, Request};
+use poem::{
+
+    Route,
+};
+use poem_openapi::{OperationId};
+use std::{sync::Arc};
+
+/// Log all requests, with important tracing data.
+pub async fn log_requests(ep: Arc<OpenTelemetryMetricsEndpoint<CorsEndpoint<Route>>>, req: Request) {
```
Mr-Leshiy, line 18: Maybe we also need to track latency of the request ?

services/voting-node/poetry.toml
stevenj: It is until I work out why the voting node won't build for me.
I thought this might be causing it,  it's doesn't seem to, so I may revert it in a later push.

stevenj:
Do not enable automerge.

containers/dev-local/docker-compose.yml
```diff
@@ -76,12 +201,29 @@ services:
...
       - leader0_store:/node_storage
       - leader0_store:/tmp/snapshot-output
 
+    # Send all logs to our internal fluentbit service.
+    logging:
+      driver: fluentd
+      options:
+        fluentd-address: ${FLUENT_BIT_HOST}:${FLUENT_BIT_PORT}
+
+  # Secondary Voting Node - 1 - TODO
```
minikin, line 216: Perhaps we could consider implementing this type of TODO system. : `TODO(username): message, URL(GitHub or Jira issue)`

src/cat-data-service/src/axum_service/mod.rs
```diff
@@ -0,0 +1,168 @@
+//! Main entrypoint to the AXUM version of the service
```
minikin, line 1: Shouldn't we explicitly call this `legacy_service` instead of `axum_service`?

minikin (changes requested):

src/cat-data-service/src/service/api/health/started_get.rs
```diff
@@ -0,0 +1,42 @@
+//! Implementation of the GET /health/started endpoint
+
+use crate::service::generic::responses::resp_2xx::NoContent;
+use crate::service::generic::responses::resp_5xx::{ServerError, ServiceUnavailable};
+
+use poem_extensions::response;
+use poem_extensions::UniResponse::T204;
+
+pub(crate) type AllResponses = response! {
```
Mr-Leshiy, line 9: I think we can move type allias to the root of the `health` module. So we can remove duplication of the code

src/cat-data-service/src/service/generic/responses/resp_5xx.rs
```diff
@@ -0,0 +1,75 @@
...
+use uuid::Uuid;
+
+use crate::settings::generate_github_issue_url;
+
+#[derive(Debug, Object)]
+#[oai(example, skip_serializing_if_is_none)]
+/// Response payload to a Bad request.
+pub(crate) struct ServerErrorPayload {
+    /// Unique ID of this Server Error so that it can be located easily for debugging.
+    pub id: Uuid,
```
Mr-Leshiy, line 18: remove `pub`

Mr-Leshiy, line 18:
```suggestion
    id: Uuid,
```

src/cat-data-service/src/service/utilities/middleware/tracing_mw.rs
```diff
@@ -0,0 +1,82 @@
...
+            version = ?req.version(),
+            method = %req.method(),
+            uri = %req.original_uri(),
+        );
+
+        if let Some(path_pattern) = req.data::<PathPattern>() {
+            span.record("path_pattern", path_pattern.0.as_ref());
+        }
+
+        async move {
```
Mr-Leshiy, line 53: What is the advantage of using tracing Span here ?

src/cat-data-service/src/settings.rs
```diff
@@ -26,3 +58,288 @@ pub struct Settings {
...
+pub(crate) fn get_api_hostnames(addr: &SocketAddr) -> Vec<String> {
+    string_to_api_hostnames(addr, API_HOSTNAMES.as_str())
+}
+
+// Jorm cleanup timeout is only used if feature is enabled.
+#[cfg(feature = "jorm-mock")]
+lazy_static! {
+    /// The jorm mock timeout, only used if feature is enabled.
+    pub(crate) static ref JORM_CLEANUP_TIMEOUT: Duration = {
+        dotenv().ok();
```
Mr-Leshiy, line 212: Why do we need that, it makes no effect.
Or I am missing something

src/cat-data-service/src/service/api/health/started_get.rs
stevenj: It's not guaranteed to be universal per endpoint; these are the responses generated by this endpoint only.
The fact they are currently all the same is a coincidence.

I would rather have a pattern where the endpoint defines everything about it and not have special cases.

src/cat-data-service/src/axum_service/mod.rs
stevenj: Ok

src/cat-data-service/src/service/generic/mod.rs
```diff
@@ -0,0 +1,5 @@
+//! Define generic reusable api components here.
```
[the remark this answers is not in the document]
stevenj: `common` is probably better.

src/cat-data-service/src/service/api/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+/// The name of the API
+const API_TITLE: &str = "Catalyst Data Service";
+
+/// The version of the API
+const API_VERSION: &str = "1.2.0";
+
+/// Get the contact details for inquiring about the API
+fn get_api_contact() -> ContactObject {
+    ContactObject::new()
+        .name("Project Catalyst Team")
```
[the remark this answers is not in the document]
stevenj: Not possible as far as I can tell because `ContactObject::new` needs to be used to create a Contact Object, which you can't do on a const.

The individual values of the fields are not exported or re-used anywhere, so I don't see a lot of value in defining:

`const API_CONTACT_NAME: &str = "Project Catalyst Team";`

and then

```
ContactObject::new().name(API_CONTACT_NAME)...
```
It seems needlessly verbose.

containers/dev-local/docker-compose.yml
minikin: We can probably introduce a custom clippy rule to automate it.

src/cat-data-service/src/service/utilities/middleware/tracing_mw.rs
stevenj: Tracing spans accumulate data.
They also log within a spans context.  So an `info!` or `warn!` inside the span shows the data it currently knows from the span and allows you to identify why the log occurred and its context without passing the context all over the place and redundantly placing it in every log message.

Example of logging with the span enabled (these are single lines in the logs, I formatted for readability):

The error log was just `error!("liveness check - error")`, but the span lets the two logs be associated.
A more practical example, if you have a `warn!` inside an `eventdb` method that is used from multiple endpoints,  the spans will let you know precisely what endpoint caused it.

Because spans can be nested,  I have set up the logs to show the nested spans, so all data from any nested span is captured, no matter where the log occurs.

Key pieces of data here:
* `conn` field which allows all logs from a single connection to be associated with each other.
* `cleint` allows all logs caused by the same client IP (anonymized) to be associated with each other.  So we can also show larger sessions of connections and interactions, and filter out all other connections happening at the same time.

```json
{
    "timestamp": "2023-09-07T08:19:28.739489325Z",
    "level": "ERROR",
    "fields": {
        "message": "liveness check - error"
    },
    "target": "cat_data_service::service::api::health::live_get",
    "filename": "src/cat-data-service/src/service/api/health/live_get.rs",
    "line_number": 39,
    "span": {
        "client": "c79f5b9f-7009-46ba-8b84-f69011026a6a",
        "conn": "01H9QBK3B3P2SM00AX2TJS1A3E",
        "method": "GET",
        "path": "/api/health/live",
        "version": "HTTP/1.1",
        "name": "request"
    },
    "spans": [
        {
            "client": "c79f5b9f-7009-46ba-8b84-f69011026a6a",
            "conn": "01H9QBK3B3P2SM00AX2TJS1A3E",
            "method": "GET",
            "path": "/api/health/live",
            "version": "HTTP/1.1",
            "name": "request"
        }
    ],
    "threadName": "tokio-runtime-worker",
    "threadId": "ThreadId(2)"
}
{
    "timestamp": "2023-09-07T08:19:28.739574687Z",
    "level": "INFO",
    "fields": {
        "message": "close",
        "time.busy": "174µs",
        "time.idle": "5.11µs"
    },
    "target": "Endpoint",
    "filename": "src/cat-data-service/src/service/utilities/middleware/tracing_mw.rs",
    "line_number": 215,
    "span": {
        "client": "c79f5b9f-7009-46ba-8b84-f69011026a6a",
        "conn": "01H9QBK3B3P2SM00AX2TJS1A3E",
        "cpu_time_ms": 0.101,
        "duration_ms": 0.109,
        "endpoint": "/api/health/live",
        "method": "GET",
        "oid": "healthLive",
        "path": "/api/health/live",
        "status": 204,
        "version": "HTTP/1.1",
        "name": "request"
    },
    "spans": [],
    "threadName": "tokio-runtime-worker",
    "threadId": "ThreadId(2)"
}
```

src/cat-data-service/src/settings.rs
stevenj: I am not sure what you mean by "no effect"?
JORM_CLEANUP_TIMEOUT is read once and converted once.  any other time its needed the pre-read and converted result is just used?

containers/dev-local/docker-compose.yml
stevenj: Yes, having a uniform set of rules for TODO, etc would be awesome.  Especially if it can be linted with clippy.

src/cat-data-service/src/axum_service/mod.rs
stevenj: Done in latest push.

src/cat-data-service/src/service/generic/responses/resp_5xx.rs
stevenj: Done in latest push

containers/dev-local/docker-compose.yml
minikin: I [created an issue](https://github.com/input-output-hk/catalyst-core/issues/556) to implement such a rule. Please have a look.

src/cat-data-service/src/settings.rs
Mr-Leshiy: Ohh, sorry just read about dotenv() function

stevenj:
The last code I pushed got both the logging and prometheus metrics working properly and properly tagged.
There are more metrics I could add, and some I would like but are non trivial like the request and response sizes.
But the current set are completely usable and probably sufficient for 90+% of all use cases.

PR #539, opened 2023-08-28
chore: Enable prometheus metrics in jormungandr by default | NPG-000

saibatizoku:
# Description

Add the `prometheus-metrics` feature to the set of defaults in jormungandr.


stevenj (approved):
Thanks

PR #541, opened 2023-08-29
fix(catalyst-toolbox): Count missing voting power | NPG-0000

stevenj:
# Description

Catalyst toolbox `snapshot` and `sve-snapshot` commands were both (with different code paths) excluding individual registrations with < threshold ada voting power.

This meant that multiple registrations to the same voting key would have missing voting power if any of those individual registrations were < the threshold.  This change does that check only on the cumulative total of all voting power.

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

The `utilities/snapshot-check/snapshot-check.py` was sued to validate the voting power is properly being calculated.
The `utilities/snapshot-check/compar_snapshot.py` was used to compare a full processed snapshot with an SVE one.

Old snapshots fail to validate, new ones after these changes validate correctly.

## Checklist

- [ ] I have made corresponding changes to the documentation
- [ ] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/src/catalyst-toolbox/snapshot-lib/src/sve.rs
+++ b/src/catalyst-toolbox/snapshot-lib/src/sve.rs
@@ -16,10 +16,12 @@ pub struct Snapshot {
 }
 
 impl Snapshot {
+    #[must_use]
+    #[allow(clippy::missing_panics_doc)] // The one possible panic shouldn't happen in reality.
     pub fn new(raw_snapshot: RawSnapshot, min_stake_threshold: Value) -> (Self, usize) {
         let mut total_rejected_registrations: usize = 0;
 
-        let inner = raw_snapshot
+        let mut inner = raw_snapshot
             .0
             .into_iter()
             .filter(|r| {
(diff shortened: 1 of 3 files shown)


src/catalyst-toolbox/snapshot-lib/src/sve.rs
```diff
@@ -48,9 +49,23 @@ impl Snapshot {
                 acc
             });
 
+        // Because of multiple registrations to the same voting key,  we can only
+        // filter once all registrations for the same key are known.
+        // `min_stake_threshold` is the minimum stake for all registrations COMBINED.
+        inner.retain(|_, regs| {
```
Mr-Leshiy, line 55: Seems it is missing `total_rejected_registrations` correct incrementing in this place

Mr-Leshiy (changes requested):

src/catalyst-toolbox/snapshot-lib/src/sve.rs
stevenj: We don't do anything with that information anyway.

minikin (approved):
LGTM

minikin (approved):

Mr-Leshiy (approved):

PR #545, opened 2023-08-31
feat: Update Review's importer, remove allocation step | NPG-8040

Mr-Leshiy:
# Description

- Removed allocation step from review's importer, it will be moved as a separate tool later
- Setup test in Earthly

### DIFF
--- a/utilities/ideascale-importer/Earthfile
+++ b/utilities/ideascale-importer/Earthfile
@@ -73,6 +73,20 @@ docker:
     SAVE IMAGE --push ${registry}ideascale-importer:$tag
 
 # Run tests
-pytest:
+test:
     FROM +build
-    RUN poetry run pytest -v --doctest-modules
+
+    COPY ../../src/event-db+docker-compose/docker-compose.yml docker-compose.yml
+    WITH DOCKER \
+        --compose docker-compose.yml \
+        --pull postgres:14 \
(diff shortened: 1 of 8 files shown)


utilities/ideascale-importer/Earthfile
```diff
@@ -73,6 +73,20 @@ docker:
...
+        --pull postgres:14 \
+        --load migrations:latest=(../../containers/event-db-migrations+docker --data=test) \
+        --service migrations \
+        --allow-privileged
+        RUN --secret IDEASCALE_EMAIL \
+            --secret IDEASCALE_PASSWORD \
+            --secret IDEASCALE_API_TOKEN \
+            EVENTDB_URL="postgres://catalyst-event-dev:CHANGE_ME@localhost/CatalystEventDev" \
+            IDEASCALE_API_URL="https://temp-cardano-sandbox.ideascale.com" \
+            poetry run pytest
```
kukkok3, line 91: I think you should run this command with --no-cache as you are testing an external service @jmgilman?

jmgilman: I haven't been able to verify if this is strictly necessary yet, but it's not a bad idea to add `--no-cache` to the `RUN` command to make sure it runs every time.

Mr-Leshiy: updated

kukkok3 (approved):
LGTM

PR #580, opened 2023-09-27
feat: Implement openapi client code generator | NPG-6475

nicopado:
# Description

Implement openapi client generator to automatically generate cat-data-service client code

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] This change requires a documentation update

## How Has This Been Tested?

- [X] Navigate to the test folder and run
`earthly +generate-client`

## Checklist

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [X] Any dependent changes have been merged and published in downstream modules


tests/Earthfile
```diff
@@ -0,0 +1,27 @@
...
+    RUN apt-get install default-jre -y
+
+generate-client:
+    FROM +dependencies
+    
+    # TODO: GET RIGHT FILE FROM POEM GENERATION / BY COPY FILE IN REPO
+    ARG openapispec_file="./petstore.yaml"
+    COPY $openapispec_file .
+
+    RUN npm install @openapitools/openapi-generator-cli -g
```
kukkok3, line 23: Moving this line in the dependencies target will make it cache even when the openapispec_file changes

tests/Earthfile
```diff
@@ -0,0 +1,27 @@
...
+    RUN apt-get install gpg -y
+    RUN curl -fsSL https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg
+    RUN apt-get install nodejs -y
+    RUN apt-get install npm -y
+    RUN apt-get install default-jre -y
+
+generate-client:
+    FROM +dependencies
+    
+    # TODO: GET RIGHT FILE FROM POEM GENERATION / BY COPY FILE IN REPO
```
kukkok3, line 19: I would avoid having a todo in the code you are merging

Mr-Leshiy: Why not ? If it is not aligned for the current ticket and could be a part of the future improvements

minikin: I agree with @kukkok3 . TODOs are technical deps. We should avoid creating them.
If we fail to fully complete the task and must create a TODO, then a TODO should have an owner and GitHub ticket.

minikin (changes requested):

tests/Earthfile
nicopado: I agree in general, but afaik the required process to generate the openapi spec via poem is not completed.
@stevenj do we have a ticket for poem implementation so that I can link this TODO to that ticket?

tests/Earthfile
nicopado: Nice to know!
We can create a separate target just for this little sense to me to add this step to a dependency target for reusability.
When we get poem running these 2 first lines will be changed with the target we'll use to generate the openapi doc

nicopado:
Added reference to github issue on comment's TODO

Mr-Leshiy (approved):

minikin (approved):

PR #590, opened 2023-10-03
feat: Handle DB connection timeouts | NPG-000

saibatizoku:
# Description

Allow `cat-data-service` to run without connection to DB, logging errors and having a configurable Retry-After header of a 503 response code.

All endpoints attempt to get a connection if none is available.

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce.
Please also list any relevant details for your test configuration

- [ ] Test A
- [ ] Test B

**Test Configuration**: _if applicable_

- Firmware version:
- Hardware:
- Toolchain:
- SDK:

**List of new dependencies**: _if applicable_

Provide a list of newly added dependencies in this PR, with the description of what are they doing and why do we need them.

- Crate A: description
- Crate B: description

## Checklist

### DIFF
--- a/src/cat-data-service/Earthfile
+++ b/src/cat-data-service/Earthfile
@@ -24,15 +24,20 @@ docker:
 
 # Need to be run with the -P flag
 test:
-    FROM ../../+builder
+    FROM earthly/dind:alpine
 
     COPY ../../src/event-db+docker-compose/docker-compose.yml docker-compose.yml
     WITH DOCKER \
         --compose docker-compose.yml \
         --pull postgres:14 \
         --load migrations:latest=(../../containers/event-db-migrations+docker --data=test) \
+        --load test:latest=(../../+builder) \
         --service migrations \
         --allow-privileged
-        RUN EVENT_DB_URL="postgres://catalyst-event-dev:CHANGE_ME@localhost/CatalystEventDev" cargo test -p cat-data-service --all-features
+        RUN docker run \
+            --network default_default \
+            -e EVENT_DB_URL="postgres://catalyst-event-dev:CHANGE_ME@postgres/CatalystEventDev" \
+            test:latest \
+                cargo test -p cat-data-service --all-features
     END
 
--- a/src/cat-data-service/src/cli.rs
+++ b/src/cat-data-service/src/cli.rs
@@ -27,7 +27,8 @@ impl Cli {
                 logger::init(settings.log_level).unwrap();
 
                 let state = Arc::new(
-                    State::new(Some(settings.database_url), settings.delay_seconds).await?,
+                    State::new_with_delay(Some(settings.database_url), settings.delay_seconds)
+                        .await?,
                 );
                 set_var(
                     RETRY_AFTER_DELAY_SECONDS_ENVVAR,
--- a/src/cat-data-service/src/cli.rs
+++ b/src/cat-data-service/src/cli.rs
@@ -27,8 +27,7 @@ impl Cli {
                 logger::init(settings.log_level).unwrap();
 
                 let state = Arc::new(
-                    State::new_with_delay(Some(settings.database_url), settings.delay_seconds)
-                        .await?,
+                    State::new(Some(settings.database_url), Some(settings.delay_seconds)).await?,
                 );
 
                 // Check the schema version, if connection to DB timesout, log as warning and
--- a/src/cat-data-service/src/cli.rs
+++ b/src/cat-data-service/src/cli.rs
@@ -31,8 +31,8 @@ impl Cli {
                         .await?,
                 );
 
-                // Check the schema version, if connection to DB timesout, log as warning,
-                // otherwise, return Error.
+                // Check the schema version, if connection to DB timesout, log as warning and
+                // continue, otherwise, return Error.
                 match state.event_db.schema_version_check().await {
                     Ok(current_ver) => {
                         tracing::info!(schema_version = current_ver, "verified schema version")
--- a/src/cat-data-service/src/cli.rs
+++ b/src/cat-data-service/src/cli.rs
@@ -1,10 +1,10 @@
 use crate::{
     logger, service,
-    settings::{Settings, RETRY_AFTER_DELAY_SECONDS_ENVVAR},
+    settings::{RetryAfterParams, Settings},
     state::State,
 };
 use clap::Parser;
-use std::{env::set_var, sync::Arc};
+use std::sync::Arc;
 
 #[derive(thiserror::Error, Debug)]
 pub enum Error {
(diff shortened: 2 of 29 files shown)


src/cat-data-service/Earthfile
```diff
@@ -24,15 +24,20 @@ docker:
 
 # Need to be run with the -P flag
 test:
-    FROM ../../+builder
+    FROM earthly/dind:alpine
```
Mr-Leshiy, line 27: Why this environment ?

saibatizoku: It's the recommended best-practice https://docs.earthly.dev/best-practices#use-earthly-dind
The main reason is that if the `../../+builder` container does not have docker, Earthly will try to install it. Unfortunately, not all containers are consistently compatible with docker, so a situation like upgrading the base container from `slim-bullseye` to `slim-bookworm` breaks these tests.

saibatizoku:
@Mr-Leshiy please take a look at this, I believe that it is ready to be merged.

src/cat-data-service/src/cli.rs
```diff
@@ -22,7 +26,30 @@ impl Cli {
...
 
-                let state = Arc::new(State::new(Some(settings.database_url)).await?);
+                let state = Arc::new(
+                    State::new_with_delay(Some(settings.database_url), settings.delay_seconds)
+                        .await?,
+                );
+
+                // Check the schema version, if connection to DB timesout, log as warning,
+                // otherwise, return Error.
+                match state.event_db.schema_version_check().await {
```
Mr-Leshiy, line 36: Why you have moved this validation out from the event-db crate ?
I think this check should under the responsibility of the db itself, not the service itself ?

src/cat-data-service/src/state/mod.rs
```diff
@@ -15,12 +15,22 @@ pub struct State {
...
     //event_db_handle: Arc<ArcSwap<Option<dyn EventDbQueries>>>, // Private need to get it with a function.
     pub event_db: Arc<dyn EventDbQueries>, // This needs to be obsoleted, we want the DB to be able to be down.
+    pub delay_seconds: u64,
     #[cfg(feature = "jorm-mock")]
     pub jorm: std::sync::Mutex<jorm_mock::JormState>,
 }
 
 impl State {
+    #[allow(dead_code)]
     pub async fn new(database_url: Option<String>) -> Result<Self, Error> {
```
Mr-Leshiy, line 25: I think it would be better to leave only one function `new` but with an argument
`delay_seconds: Option<u64>`, so if it is `None` you can initialise with the default value ` RETRY_AFTER_DELAY_SECONDS_DEFAULT`

src/cat-data-service/src/legacy_service/health.rs
```diff
@@ -1,23 +1,46 @@
...
+                move |params| async { retry_after_exec(params, state).await }
+            }),
+        )
 }
 
-async fn ready_exec() -> Result<bool, Error> {
+async fn ready_exec(state: Arc<State>) -> Result<bool, Error> {
     tracing::debug!("health ready exec");
 
+    state.event_db.schema_version_check().await?;
```
Mr-Leshiy, line 43: Same question as above

src/cat-data-service/src/legacy_service/health.rs
```diff
@@ -27,6 +50,54 @@ async fn live_exec() -> Result<bool, Error> {
     Ok(true)
 }
 
+async fn retry_after_exec(Query(params): Query<RetryAfterParams>, state: Arc<State>) -> Response {
```
Mr-Leshiy, line 53: Not quite clear why this endpoint is needed, can you provide some documentation pls

Mr-Leshiy (commented):
BTW I have realised that most of the updates related to the old `axum` implementation of the service. Do we really want to modify this part ?
I am asking because as this PR is going to be merged to the `feature/poem-migration-start` branch, so it seems that we need to provide this logic for new `poem` part.

src/cat-data-service/src/cli.rs
saibatizoku: @Mr-Leshiy I have reverted the responsibility to the db and created an alternative way for the service to run the schema version check.

src/cat-data-service/src/legacy_service/health.rs
saibatizoku: In this case, the `/health/ready` endpoint should check that the connection to the database is working as expected. The simplest way that I could think of to implement this with existing functionality is to verify the schema version.

src/cat-data-service/src/legacy_service/health.rs
saibatizoku: This endpoint is used to configure the `retry-after` header, that can be a datetime string or a positive integer, in 503 responses. This is useful to inform other services when they should expect to retry connection to our service.

src/cat-data-service/src/state/mod.rs
saibatizoku: @Mr-Leshiy I've done as you suggested.

src/cat-data-service/src/service/api/health/ready_get.rs
```diff
@@ -35,7 +40,22 @@ pub(crate) type AllResponses = response! {
 /// * 500 Server Error - If anything within this function fails unexpectedly. (Possible but unlikely)
 /// * 503 Service Unavailable - Service is not ready, do not send other requests.
 #[allow(clippy::unused_async)]
```
Mr-Leshiy, line 42:
```suggestion
```

src/cat-data-service/src/service/api/health/ready_get.rs
```diff
@@ -35,7 +40,22 @@ pub(crate) type AllResponses = response! {
...
-pub(crate) async fn endpoint() -> AllResponses {
+pub(crate) async fn endpoint(state: Data<&Arc<State>>) -> AllResponses {
     // otherwise everything seems to be A-OK
-    T204(NoContent)
+    match state.event_db.schema_version_check().await {
```
Mr-Leshiy, line 45: Seems like it makes sense to extract this code to the separate function, because it is used more than one place

src/cat-data-service/src/state/mod.rs
```diff
@@ -15,16 +15,23 @@ pub struct State {
...
     //event_db_handle: Arc<ArcSwap<Option<dyn EventDbQueries>>>, // Private need to get it with a function.
     pub event_db: Arc<dyn EventDbQueries>, // This needs to be obsoleted, we want the DB to be able to be down.
+    pub delay_seconds: u64,
     #[cfg(feature = "jorm-mock")]
     pub jorm: std::sync::Mutex<jorm_mock::JormState>,
 }
 
 impl State {
-    pub async fn new(database_url: Option<String>) -> Result<Self, Error> {
+    #[allow(dead_code)]
```
Mr-Leshiy, line 24:
```suggestion
```

Mr-Leshiy (approved):
Despite few comments LGTM

src/cat-data-service/src/service/api/health/ready_get.rs
saibatizoku: I agree. At a first attempt, I realize that I need to learn better how to make use of poem-extensions to move this into a cleaner way of handling 5xx responses. If you don't mind, @Mr-Leshiy , this could be done in a separate PR.

Mr-Leshiy (approved):

PR #596, opened 2023-10-12
feat: Fragment generator | NPG-0000

cong-or:
Fragment generator in byte format. [As per specification](https://github.com/input-output-hk/catalyst-core/blob/main/src/chain-libs/chain-impl-mockchain/doc/format.abnf).

The original idea was to isolate and rewrite a bare bones version of the specification from scratch to verify the spec and our existing implementation. This was largely done except for the cryptography. The encrypted vote part is [manageable to rewrite](https://github.com/input-output-hk/catalyst-core/blob/main/src/chain-libs/chain-impl-mockchain/doc/format.abnf#L170) and somewhat **standard**. 

However the ZKP proof is not, the algorithm from the treasury paper is NOVEL and difficult to recreate with external off the shelf libs. An isolated ZKP implementation with no baggage in the context of cat-core requires a rewrite from scratch as there is no existing libs suitable; any rewrite is basically a mirror of our existing cryptography and vote crates in cat-core. 

In any case, we need to use the binary of this crate within the context of our load tester. We can potentially come back to alternatives implementation of the ZKP stuff in the future.


kukkok3 (approved):
LGTM

PR #624, opened 2023-11-15
chore(fix-stage-data-sql): Upsert stage data | NPG-0000

### LINKED ISSUE #500: Stage data does not cleanup before insert
The new stage data does not do any cleanup prior to the `INSERT`. In cases where migrations are being deployed, and the stage data hasn't changed the event number (i.e. 99% of the deployments), the migration will fail with:

```
Adding stage specific data from ./stage_data/testnet/00001_fund10_event.sql
psql:./stage_data/testnet/00001_fund10_event.sql:54: ERROR:  duplicate key value violates unique constraint "event_pkey"
DETAIL:  Key (row_id)=(10) already exists.
Adding stage specific data from ./stage_data/testnet/00002_fund10_ideascale_params.sql
psql:./stage_data/testnet/00002_fund10_ideascale_params.sql:10: ERROR:  duplicate key value violates unique constraint "config_idx"
DETAIL:  Key (id, id2, id3)=(ideascale, params, F10) already exists.
psql:./stage_data/testnet/00002_fund10_ideascale_params.sql:18: ERROR:  duplicate key value violates unique constraint "config_idx"
DETAIL:  Key (id, id2, id3)=(event, ideascale_params, 10) already exists.
>>> Finished entrypoint script
```

The stage data should drop the associated row before inserting, otherwise we'll not be able to modify any of the stage data without manual intervention.

FelipeRosa:
Fix duplicate key errors by changing stage_data SQL statements to  upsert event and config data instead of inserting.

Closes #500.

### DIFF
--- a/src/event-db/stage_data/dev/00001_fund11_event.sql
+++ b/src/event-db/stage_data/dev/00001_fund11_event.sql
@@ -51,4 +51,28 @@ INSERT INTO event (
     1,                      -- Committee Threshold
     NULL,                   -- Extra
     NULL                    -- Cast to
-);
+) ON CONFLICT (row_id) DO UPDATE
+SET name = EXCLUDED.name,
+    description = EXCLUDED.description,
+    registration_snapshot_time = EXCLUDED.registration_snapshot_time,
+    snapshot_start = EXCLUDED.snapshot_start,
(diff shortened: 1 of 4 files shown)


jmgilman (approved):
LGTM.

PR #629, opened 2023-11-17
fix: crs hex::decode  | NPG-000

cong-or:
In proof systems, provers and the verifiers rely on a common set of parameters, sometimes referred to as the common reference string (CRS).

we are using the the DigestOf<Blake2b256, VotePlan>; as the CRS

align with on the fly load tester updates


FelipeRosa (approved):
LGTM

Mr-Leshiy (approved):
LGTM

PR #652, opened 2024-01-08
feat: (yay or nay): add choice to current jorm crypto config feat | NPG-000

cong-or:
add choice flag to vote signer, tied to current jorm config where 0 is NO and 1 is YES in the context of position in array.


kukkok3 (approved):
LGTM

PR #675, opened 2024-02-15
feat: build(bump to 1.75): cargo chef requires 1.75 | NPG-000

cong-or:
1.71 is causing breaking builds with outdated deps via cargo chef.


kukkok3 (approved):
LGTM

PR #690, opened 2024-04-09
fix: cleans up eventdb entrypoint and adds fix for RDS | NPG-000

jmgilman:
This PR addresses the following issues:

1. We no longer have a dependency on graphql, so it removes all init code related to that in the eventdb entrypoint
1. The entrypoint was trying to connect to a database that didn't exist
   a. It was changed to connect to the "root" database on initialization (a new input was added for this) 
1. For initialization to work with RDS, the root role must be assigned the role the database is being created with
   a. This is a weird RDS requirement where the root role isn't really "super"
1. The debug print statements in the init script were printing out the raw database password every single run
   a. Even though our logs are private, this is still a serious issue
   b. The debug statements were moved to the entrypoint instead so they could be conditionally controlled
1. The ideascale SQL for the dev environment had a syntax error in it

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -226,7 +226,7 @@ jobs:
         uses: taiki-e/install-action@nextest
 
       - name: Install cargo-make
-        run: cargo install --force cargo-make --locked
+        run: cargo install --force cargo-make --version 0.37.11 --locked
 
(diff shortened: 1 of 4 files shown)


kukkok3 (approved):
LGTM

PR #727, opened 2024-12-10
fix: Include necessary deps for event-db-migrations build.

coire1:
# Description

Add deps to fix eventdb-migrations build.


kukkok3 (approved):
LGTM

PR #728, opened 2024-12-11
fix: Bump refinery version | NPG-000

coire1:
# Description

Bump the refinery version to fix the build in catalyst-reviews.


kukkok3 (approved):
LGTM

PR #768, opened 2025-10-06
feat: Apply gamma to single voter to calculate total stake threshold | NPG-000

coire1:
Apply gamma to single voter to calculate total stake threshold.


cong-or (approved):
LGTM

